PR #57, opened 2017-05-15
replacing user_token with project_token

phillip-nguyen-work:
### Purpose/goal of Pull Request (Issue #)
For clarity in the communication with user, we need to use the name project token everywhere instead of user token
Related to: https://github.com/WOVNio/equalizer/issues/4057

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Settings.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Settings.java
@@ -12,7 +12,7 @@ class Settings {
     static final String UrlPatternRegQuery = "(?:(?:\\?.*&)|\\?)wovn=([^&]+)(?:&|$)";
     static final String UrlPatternRegSubdomain = "^([^.]+)\\.";
 
-    String userToken = "";
+    String projectToken = "";
     String secretKey = "";
     String urlPattern = "path";
     String urlPatternReg = UrlPatternRegPath;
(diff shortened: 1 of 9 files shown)


phillip-nguyen-work:
@a-bougouin Let me check again using the user token. In my initial testing, I had used the project token and it was translating.

src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -37,6 +37,11 @@
 
         String p;
 
+        p = config.getInitParameter("userToken");
+        if (p != null && p.length() > 0) {
+            this.projectToken = p;
```
a-bougouin: if both userToken and projectToken are in configuration, then projectToken should not be replaced. Imo you miss a case in your if statement.

phillip-nguyen-work: this.projectToken is how it's stored in the code.

The section below it ensures that projectToken takes priority, because if it's configured, it will replace userToken.

a-bougouin: I see, thanks for the details.
Do you have a test case when userToken is set and projectToken is set with a different token? Please add one if not.

phillip-nguyen-work: I added a test to check that projectToken takes priority over projectToken

phillip-nguyen-work: @a-bougouin can you take a look at the changes?

a-bougouin: sorry for late notice

PR #58, opened 2017-11-20
Keep doctype when translate html

hythof:
### Purpose/goal of PR (Issue #6792)
Relate: https://github.com/WOVNio/equalizer/issues/6792

Keep doctype when translate html.

Because `HtmlDocumentBuilder` that library remove doctype when create DOM tree.

### Comments
Update JDK version in .travis.yml

- Add Open JDK 8  
  Because Oracle JDK 8 is supported.
- Remove Oracle JDK 7  
  Because [that version of the JDK is no longer current ](http://www.oracle.com/technetwork/java/javase/downloads/jdk7-downloads-1880260.html)
- Remove Open JDK 6  
  Because [Sun/Oracle JDK 6 is not provided because it reached End of Life in fall 2012.](https://docs.travis-ci.com/user/reference/precise/)

I confirmed jeff about remove JDK6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,5 @@ language: java
 
 jdk:
   - oraclejdk8
-  - oraclejdk7
+  - openjdk8
   - openjdk7
-  - openjdk6
-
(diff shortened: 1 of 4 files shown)


a-bougouin:
If there is new doctype versions, it won't work unless we update the code.
Did you do that do avoid detecting the doctype by regular expression?

hythof:
The idea sounds good.
I changed to a method that using regular expression.

PR #60, opened 2017-12-07
Hotfix/punycode for meta tag

hythof:
### Purpose/goal of PR (Issue #7034)
https://github.com/WOVNio/equalizer/issues/7034

Change tag attribute to punycode if it include internationalized domain name.

The list of attributes.
```
- <link href="...">
- <a href="...">
- <script src="...">
- <img src="...">
- <firame src="...">
- <meta content="..." name="og:url | og:image | twitter:url | twitter:image">
```

### Comments
I confirmed this solution to Shoji about using punycode.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Interceptor.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Interceptor.java
@@ -333,6 +333,7 @@ class Interceptor {
         changeUrlToPunyCode(doc, "a", "href");
         changeUrlToPunyCode(doc, "script", "src");
         changeUrlToPunyCode(doc, "img", "src");
+        changeUrlToPunyCode(doc, "iframe", "src");
 
         if (!lang.equals(this.store.settings.defaultLang)) {
             NodeList anchors = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/wovnio/wovnjava/Interceptor.java
```diff
@@ -507,4 +511,65 @@ private String switchLang(String body, Values values, HashMap<String, String> ur
...
+
+    private final Pattern extractDomain = Pattern.compile("(?:https?:)?//([^/]+)");
+
+    private void changeUrlToPunyCode(Document doc) {
+        changeUrlToPunyCodeAttr(doc, "link", "href");
+        changeUrlToPunyCodeAttr(doc, "a", "href");
+        changeUrlToPunyCodeAttr(doc, "script", "src");
+        changeUrlToPunyCodeAttr(doc, "img", "src");
+        changeUrlToPunyCodeAttr(doc, "iframe", "src");
+        changeUrlToPunyCodeAttr(doc, "meta", "content");
```
suplo: might need to add `changeUrlToPunyCodeAttr(doc, "input", "src");`

hythof: That's right.
I added it.
Thanks.

PR #61, opened 2018-02-19
Hotfix/feature site prefix path

hythof:
Create `site_prefix_path` feature.

I explain this feature.
Normally, WOVN change URL `/global/index.html`  to `/fr/global/index.html`.
If user put `sitePrefixPath` and `/global` to settings, WOVN change URL `/global/index.html`  to `/global/fr/index.html`.


src/test/java/com/github/wovnio/wovnjava/HeadersTest.java
```diff
@@ -350,15 +378,16 @@ public void testServerPort() {
         assertEquals("example.com:8080/test", h.pageUrl);
     }
 
-    public void testOriginalHeaders() {
```
mrasu: why removed?

hythof: Just a mistake.
I restored it.

PR #67, opened 2018-05-25
Disable pretty print

hythof:
Relate: https://github.com/WOVNio/equalizer/issues/9799

Disable pretty print for reduce html size.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
@@ -18,7 +18,7 @@ class HtmlConverter {
     HtmlConverter(Settings settings, String original) {
         this.settings = settings;
         doc = Jsoup.parse(original);
-        doc.outputSettings().indentAmount(0).prettyPrint(false);
+        doc.outputSettings().prettyPrint(false);
     }
 
     String strip() {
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -18,22 +18,23 @@
     HtmlConverter(Settings settings, String original) {
         this.settings = settings;
         doc = Jsoup.parse(original);
+        doc.outputSettings().indentAmount(0).prettyPrint(false);
```
a-bougouin: Is it needed to set `indentAmount`? Isn't it used for pretty printing only?

a-bougouin:
Could you have a test that asserts original HTML indenting is preserved?

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
hythof: I'm wrong. The option is for pretty printing only.
I removed it.
Thanks.

hythof:
I added a test for pretty printing.

src/test/java/com/github/wovnio/wovnjava/HtmlConverterTest.java
```diff
@@ -10,6 +10,14 @@
 
 public class HtmlConverterTest extends TestCase {
 
+    public void testDisablePrettyPrint() {
+        String original = "<html><head></head><body>\n" + "hello" + "\n</body></html>";
```
a-bougouin: Could you add spacing before "hello"?:
```
String original = "<html><head></head><body>\n" + "  hello" + "\n</body></html>";
```

PR #73, opened 2018-07-25
Hotfix/include url schema

torkelbe:
Include schema("http://") in request to API server.

#### Problem
Related to https://github.com/WOVNio/equalizer/issues/9799

wovnjava would send `url=example.com` to html-swapper, and html-swapper then changed it to `://example.com`, causing the error.

#### Solution
Now we make sure to always send `url=http[s]://example.com`.

This change includes some reorganization in the relevant tests in order to properly test the request that gets sent to the API server.

It also includes a version bump to 0.3.7

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -183,6 +183,9 @@ class Headers {
         return this.pathLang;
     }
 
+    /**
+     * @return String Returns request URL with new language code added
+     */
     String redirectLocation(String lang) {
         if (lang.equals(this.settings.defaultLang)) {
             return this.protocol + "://" + this.url;
@@ -206,6 +209,13 @@ class Headers {
             }
             return protocol + "://" + location;
         }
+    }
+
+    /**
+     * @return String Returns request URL without any language code
+     */
+    String getUrlWithoutLanguageCode() {
+        return this.protocol + "://" + this.url;
     }
 
     String removeLang(String uri, String lang) {
--- a/src/test/java/com/github/wovnio/wovnjava/ApiTest.java
+++ b/src/test/java/com/github/wovnio/wovnjava/ApiTest.java
@@ -2,6 +2,7 @@ package com.github.wovnio.wovnjava;
 
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.BufferedReader;
 import java.io.ByteArrayOutputStream;
 import java.io.ByteArrayInputStream;
 import java.io.UnsupportedEncodingException;
@@ -13,6 +14,7 @@ import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 import java.net.HttpURLConnection;
 import java.net.ProtocolException;
+import java.net.URLDecoder;
 import javax.servlet.FilterConfig;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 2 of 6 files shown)


src/test/java/com/github/wovnio/wovnjava/ApiTest.java
```diff
@@ -66,18 +87,32 @@ private String translate(HttpServletRequest request, Settings settings, String h
...
+    private static String decompress(byte[] compressed) throws IOException {
+        ByteArrayInputStream bis = new ByteArrayInputStream(compressed);
+        GZIPInputStream gis = new GZIPInputStream(bis);
+        BufferedReader br = new BufferedReader(new InputStreamReader(gis, "UTF-8"));
+        StringBuilder sb = new StringBuilder();
+        String line;
+        while((line = br.readLine()) != null) {
+            sb.append(line);
+        }
+        br.close();
```
kubo39: Don't we have to care about exceptional safety?

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -208,6 +211,13 @@ String redirectLocation(String lang) {
         }
     }
 
+    /**
+     * @return String Returns request URL without any language code
+     */
+    String getUrlWithoutLanguageCode() {
```
kubo39: confusing...

src/test/java/com/github/wovnio/wovnjava/ApiTest.java
```diff
@@ -66,18 +87,32 @@ private String translate(HttpServletRequest request, Settings settings, String h
...
+        EasyMock.expect(mock.getOutputStream()).andReturn(requestStream);
         EasyMock.expect(mock.getInputStream()).andReturn(responseStream);
         EasyMock.replay(mock);
         return mock;
     }
+
+    private static String decompress(byte[] compressed) throws IOException {
+        ByteArrayInputStream bis = new ByteArrayInputStream(compressed);
+        GZIPInputStream gis = new GZIPInputStream(bis);
+        BufferedReader br = new BufferedReader(new InputStreamReader(gis, "UTF-8"));
```
kubo39: Nit: `java.util.Scanner` or `java.util.nio.File` seems better for me, but don't have to change.

src/test/java/com/github/wovnio/wovnjava/ApiTest.java
hythof: I think that it's okay because this is test.

src/main/java/com/github/wovnio/wovnjava/Headers.java
hythof: Looks good.

src/test/java/com/github/wovnio/wovnjava/ApiTest.java
torkelbe: Thanks, I'll close the streams in a `finally` clause.

src/test/java/com/github/wovnio/wovnjava/ApiTest.java
torkelbe: I don't have a preference, so I'll leave it like this unless I get any further objections.

src/main/java/com/github/wovnio/wovnjava/Headers.java
kubo39: I see.

torkelbe: Alright, lets make it less confusing.

At first I named it `originalLocation`, as opposed to the `redirectLocation` function above it. But that felt imprecise. So I changed it to the current name because we use the function when we want to read the request URL, just stripped of any language code. (Similarly, the function above is used for when we want the request URL with the language code included.)

Do you have a suggestion for how we can name it better?

kubo39: First I expected `this.url` includes scheme, because [rfc says](https://tools.ietf.org/html/rfc3986).

torkelbe: Yes, I think that the original author may have expected the same thing. And since the tests did not cover the request format, it resulted in this bug.

PR #75, opened 2018-08-02
Support location correctly without test

hythof:
Relate: https://github.com/WOVNio/equalizer/issues/11477
Replace `Location` header correctly.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -211,6 +211,62 @@ class Headers {
         }
     }
 
+    public String locationWithLangCode(String location) {
+        // check if needed
+        if (location == null) {
+            return null;
+        }
+        if (langCode().equals(settings.defaultLang)) {
+            return location;
+        }
+
+        // catprue protocl and path
+        String locationProtocol = protocol;
+        String path;
+        if (location.contains("://")) {
+            if (!location.contains("://" + host)) {
+                return location;
+            }
+            String[] protocolAndRemaining = location.split("://", 2);
+            locationProtocol = protocolAndRemaining[0];
+            path = "/" + protocolAndRemaining[1].split("/", 2)[1];
+        } else {
+            path = location.startsWith("/") ? location : pathJoin("/", pathJoin(pathNameKeepTrailingSlash, location));
+        }
+        path = pathNormalize(path);
+
+        // check location already have language code
+        if (settings.urlPattern.equals("query") && location.contains("wovn=")) {
+            return location;
+        } else if (settings.urlPattern.equals("path")) {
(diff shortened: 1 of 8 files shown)


src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -211,6 +211,62 @@ String redirectLocation(String lang) {
...
+            } else {
+                queryLangCode = "?wovn=" + lang;
+            }
+        } else if (settings.urlPattern.equals("subdomain")) {
+            subdomainLangCode = lang + ".";
+        } else {
+            pathLangCode = "/" + lang;
+        }
+        return locationProtocol + "://" + subdomainLangCode + host + pathLangCode + path + queryLangCode;
+    }
```
torkelbe: What is the difference between
```
String redirectLocation(String lang) {...}
```
and
```
public String locationWithLangCode(String location) {...}
```
To me, they seem to try to do the same thing.

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -234,21 +290,33 @@ String removeLang(String uri, String lang) {
...
-                if (Pattern.compile("\\?").matcher(loc).find()) {
-                    res.setHeader("Location", loc + "?wovn=" + langCode());
-                } else {
-                    res.setHeader("Location", loc + "&wovn=" + langCode());
-                }
-            } else if (this.settings.urlPattern.equals("subdomain")) {
-                loc = loc.replaceFirst("//([^.]+)", "//" + langCode() + "\\.$1");
-                res.setHeader("Location", loc);
+    private String pathJoin(String left, String right) {
+        boolean l = left.endsWith("/");
```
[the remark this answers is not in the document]
hythof: If left is "foo" and right is "./bar", the result is "foo/./bar".
After that it's converted "foo/./bar" to "foo/bar" by `pathNormalize`.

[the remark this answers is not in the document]
hythof: If left is "foo" and right is "./bar", the result is "foo/./bar".
After that it's converted "foo/./bar" to "foo/bar" by `pathNormalize`.

src/main/java/com/github/wovnio/wovnjava/Headers.java
hythof: > redirectLocation
This method is to create new URL from browser URL and specific language code.
borwser URL: https://foo.com/
language code: fr
result: https://foo.com/fr/

> locationWithLangCode
This method is to create new URL from specific URL and current language code.
borwser URL: https://foo.com/fr/signin
specific URL: /home
result: https://foo.com/fr/home

torkelbe: Gotcha. Thanks for the examples.

PR #83, opened 2019-02-07
Disable super.flushBuffer on response wrapper

torkelbe:
### Comments
If `flushBuffer()` gets called on the ServletResponse, it commits changes to the response such that we can no longer change headers. However, wovnjava relies on changing `content-length` because we alter the size of the response.

Here is from [Oracle docs](https://docs.oracle.com/javaee/7/api/javax/servlet/ServletResponse.html#flushBuffer--):
```
void flushBuffer()
          throws IOException
Forces any content in the buffer to be written to the client. A call to this method
automatically commits the response, meaning the status code and headers will be written.
```

Using the library with a servlet deployed on WebLogic, the response was flushed before reaching our filter, which made us unable to change `content-length` to a bigger value. Therefore, the last part of the document was lost.

### Request before change
Request for original language receives full document, while the request for translated language loses the last part of its document. (content-length remains at 637)
```
$ curl -v weblogic.test/encode/utf.html
...
< HTTP/1.1 200 OK
< date: Thu, 07 Feb 2019 06:22:26 GMT
< accept-ranges: bytes
< content-length: 637
< content-type: text/html; charset=utf-8
...
```
```
$ curl -v weblogic.test/en/encode/utf.html
...
< HTTP/1.1 200 OK
< date: Thu, 07 Feb 2019 06:33:07 GMT
< accept-ranges: bytes
< content-length: 637
< content-type: text/html; charset=utf-8
...
```

### Request after change
After the change, the request for translated language is able to expand content length, and the full document is received. (content-length increased to 853)
```
$ curl -v weblogic.test/en/encode/utf.html
...
< HTTP/1.1 200 OK
< date: Thu, 07 Feb 2019 06:59:24 GMT
< accept-ranges: bytes
< content-length: 853
< content-type: text/html; charset=utf-8
...
```

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java
+++ b/src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java
@@ -17,6 +17,8 @@ class WovnHttpServletResponse extends HttpServletResponseWrapper {
     private ServletOutputStream output;
     private Headers headers;
 
+    private boolean disableFlushBuffer = true;
+
     WovnHttpServletResponse(HttpServletResponse response, Headers headers) {
         super(response);
         this.buff = new ByteArrayOutputStream();
@@ -79,5 +81,28 @@ class WovnHttpServletResponse extends HttpServletResponseWrapper {
             value = headers.locationWithLangCode(value);
         }
         super.setHeader(name, value);
+    }
+
+    public void flushBuffer() throws IOException {
+        flush();
+
+        // Calling `super.flushBuffer` may lead to response already being committed.
+        // This prevents us from, among other things, changing the HTTP content-length.
+        // This problem happens in Weblogic when a servlet forward method trigger a
+        // flush before it forwarded.
+        // disableFlushBuffer for that purpose is 'true' by default
+        // See: https://jira.terracotta.org/jira/si/jira.issueviews:issue-html/EHC-447/EHC-447.html
+        if (!disableFlushBuffer) {
+          super.flushBuffer();
+        }
+    }
+
+    public void flush() throws IOException {
+        if (writer != null) {
+            writer.flush();
+        }
+        if (output != null) {
+            output.flush();
+        }
     }
 }
--- a/src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java
+++ b/src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java
@@ -17,8 +17,6 @@ class WovnHttpServletResponse extends HttpServletResponseWrapper {
     private ServletOutputStream output;
     private Headers headers;
 
-    private boolean disableFlushBuffer = true;
-
     WovnHttpServletResponse(HttpServletResponse response, Headers headers) {
         super(response);
         this.buff = new ByteArrayOutputStream();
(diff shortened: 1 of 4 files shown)


hythof:
@torkelbe I think that we need always ignore `flushBuffer`.
Because if do it, HTTP response send to client before we translated.

If we support `flushBuffer`, I think that it's better to call real `flushBuffer` after translation.

torkelbe:
Do you think we need to support `flushBuffer`?

I'm not sure in what situations it would make sense to be calling it manually. Personally, I think that it is ok for us to disable `super.flushBuffer`, and not call it at all.

I added the variable `disableFlushBuffer` to make it clear in the code what we are doing with the function (so we don't "silently" disable it).

The other function `flush()` may not be necessary, though. My thinking was that we can flush the output buffers that we have, without committing the header etc.

hythof:
> I'm not sure in what situations it would make sense to be calling it manually.
It's useful to specific situation.

For example.
```
void sendMails() {
  out.println("<html>...</html>");

  res.flushBuffer();

  // It's take time so it's better to flush html before into heavy processing
  for (i=0; i<99999999999; ++i) {
    sendmail(users[i]);
  }
}
```

torkelbe:
@hythof One option we talked about was to handle translation immediately when flushBuffer is called on the wrapped ServletResponse, kind of like this
```java
void flushBuffer() {
  translatedBody = tryTranslate(wovnResponse.body)
  response.write(translatedBody)
  response.flushBuffer()
}
```
But flushBuffer can be called before the entire response is ready, in which case we would be attempting to translate only part of the document. I think that's not what we want.

I cannot find a specification for using flushBuffer in a filter, but it seems like the normal procedure is to call flush on the output buffers in the wrapper, and wait with committing the response until `chain.doFilter` returns.

One thing we could do is to call flushBuffer on the response object before WovnServletFilter is finished (if flushBuffer had been called on our response wrapper). That is now added in a new commit. At least it propagates the flushBuffer call, but it still waits with writing anything to the client until all the content is ready. I don't think we can really get around this, however.

torkelbe:
**Note:** from https://docs.oracle.com/middleware/1213/wls/WBAPP/bestpractices.htm#WBAPP640
> ### Best Practice When Subclassing ServletResponseWrapper
> Java EE provides the class `javax.servlet.ServletResponseWrapper`, which you can subclass in your Servlet to adapt its response.
> Oracle recommends that if you create your own response wrapper by subclassing the ServletResponseWrapper class, you should always override the `flushBuffer()` and `resetBuffer()` methods. Not doing so might result in the response being committed prematurely.

That's the closest to an official guideline I found, but it leaves out details on how `flushBuffer()` is expected to propagate.

torkelbe:
The change looks like it is successful for the customer using it, and I would like to commit it.

Calls to flushBuffer is no longer propagated to the original response instance. The change enables us to modify HTTP headers in the response, and it has removed a vulnerability to errors such as IllegalStateException when attempting to access `response.getWriter()` and `response.getOutputBuffer()`.

src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java
```diff
@@ -80,4 +80,25 @@ public void setHeader(String name, String value) {
         }
         super.setHeader(name, value);
     }
+
+    public void flushBuffer() throws IOException {
+        flush();
```
hythof: Call `flush()` is need before call `super.flushBuffer()`.

But when disabled `super.flushBuffer()`,
that's not needed.
How about moving `flush()` to into `if (this.headers.settings.enableFlushBuffer) {`?

torkelbe: > But when disabled `super.flushBuffer()`, that's not needed.

Is this true though?

The behavior of `writer.flush()` and `output.flush()` depends on the implementation of the underlying stream. In our particular case it actually does nothing, because the underlying stream is ultimately an `OutputStream`.

But if the implementation of `FilterServletOutputStream` was different, and flush wrote buffered content, I think it would make sense to do that whenever the wrapped servlet calls `flushBuffer()`.

Alternatively, I can actually remove `flush()` completely.

hythof: Of course, currently `flush()` do nothing. So no problem.

I just concern below behavior could be a problem in the future.

Programmer can guess why nothing happen when call `flushBuffer()` because of `enableFlushBuffer` is disabled. But actually, call `flush()` and does not call `super.flushBuffer()`.

It's also hard to guessing without code reading. 


But it's not big thing. I approve  it even if no change.

PR #84, opened 2019-03-19
Catch flushBuffer on response wrapper

torkelbe:
This change makes sure that when `response.flushBuffer()` is called on the WovnHttpServletResponse, the call is not propagated to the wrapped original ServletResponse.

If flushBuffer is called on the original ServletResponse, we can no longer modify headers to f.ex. increase content-length (to have room for translated content). The filter also becomes vulnerable to errors such as IllegalStateException when attempting to access `response.getWriter()` and `response.getOutputBuffer()`.

See also the initial flushBuffer PR for discussion on the change: https://github.com/WOVNio/wovnjava/pull/83
That PR was used for investigating erroneous behavior on the customer's server, and will not be merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,3 +211,11 @@ Including three classes, `email-address-element`, `my-secret-class`, and `noshow
   ...
 </filter>
 ```
+
+### 2.10. enableFlushBuffer
+This parameter is set to `false` by default.
+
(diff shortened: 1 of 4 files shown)


torkelbe:
Closing this in favor of the original flushBuffer PR. Sorry.

PR #85, opened 2019-03-21
Create feature sitePrefixPath

torkelbe:
sitePrefixPath lets the customer set a prefix path to use as an anchor for which WOVN will translate pages. With this setting, WOVN will only translate pages that match the prefix path, and the path language code will be added after the prefix path.

With this setting the following behaviors are affected
* Only process requests that match prefix path
* Detect language code in request path after matching prefix path
* Insert hreflangs with language code after matching prefix path
* Send sitePrefixPath setting in snippet settings (if the setting is used)
* Send sitePrefixPath setting in request to translation API (always)

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -311,6 +311,10 @@ class Headers {
         }
     }
 
+    boolean isValidPath() {
+        return this.pathName.startsWith(this.settings.sitePrefixPathWithoutSlash);
+    }
+
     private String removeFilePart(String path) {
         if (path.endsWith("/")) {
             return path;
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -201,10 +201,20 @@ class Headers {
                 location = lang.toLowerCase() + "." + location;
             } else {
                 // path
-                if (location.contains("/")) {
-                    location = location.replaceFirst("/", "/" + lang + "/");
+                if (settings.hasSitePrefixPath) {
+                    String sitePrefixPath = this.settings.sitePrefixPathWithoutSlash;
+                    if (this.pathName.startsWith(sitePrefixPath)) {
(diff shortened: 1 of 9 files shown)


src/test/java/com/github/wovnio/wovnjava/HeadersTest.java
```diff
@@ -248,6 +248,27 @@ public void testRedirectLocationPathNestedDirectory() {
...
+    }
+
+    public void testRedirectLocationWithSitePrefixPathMatchingPath() {
+        Headers h = makeHeaderWithSitePrefixPath("/global/tokyo/", "/global/");
+        assertEquals("https://example.com/global/tokyo/", h.redirectLocation("en")); // `en` is defaultLang
+        assertEquals("https://example.com/global/ja/tokyo/", h.redirectLocation("ja"));
+        assertEquals("https://example.com/global/garbage/tokyo/", h.redirectLocation("garbage"));
+    }
+
+    public void testRedirectLocationWithSitePrefixPathNonmatchingPath() {
```
bstewart00: Here's how I would name these tests that test a single scenario:
`redirectLocation_WithSitePrefixPathBasePathOnly_AddsLanguageToUrl`
`redirectLocation_WithSitePrefixPathMatchingPath_AddsLanguageToUrl`
`redirectLocation_WithSitePrefixPathNonmatchingPath_DoesNotModifyUrl`

Just some food for thought! I think it helps readability
https://medium.com/@lorenzen.jacob/unit-test-naming-conventions-875786f53790
https://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -239,6 +249,10 @@ public String locationWithLangCode(String location) {
         }
         path = pathNormalize(path);
 
+        if (!path.startsWith(this.settings.sitePrefixPathWithoutSlash)) {
```
bstewart00: Will this still behave correctly when they aren't using "site prefix path"?

torkelbe: It seems to work well.

`sitePrefixPathWithoutSlash = ""` by default.
`path.startsWith("")` will always be true.

PR #87, opened 2019-03-21
Add devMode setting

torkelbe:
`devMode` will set the snippet URL and translation API URL to match development environment.

Set it as follows
```xml
<init-param>
    <param-name>devMode</param-name>
    <param-value>1</param-value>
</init-param>
```

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
@@ -131,7 +131,7 @@ class HtmlConverter {
         sb.append("&langCodeAliases={}&version=");
         sb.append(settings.version);
         String key = sb.toString();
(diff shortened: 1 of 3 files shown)


a-bougouin:
Dev mode should also send requests to local html-swapper, right? I don't see it in this PR.
Here is implementation at `wovnrb` for reference: https://github.com/WOVNio/wovnrb/pull/136

torkelbe:
>Dev mode should also send requests to local html-swapper, right? I don't see it in this PR. ...

@a-bougouin It should send to local html-swapper, but I don’t think we have an established standard for which port that is served from. An ‘apiUrl’ setting already exists for that purpose.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

torkelbe:
`devMode` now automatically sets local URLs for snippet and translation API.

PR #90, opened 2019-04-02
Set content-type to UTF-8 also for fallback method

torkelbe:
When translating by API, we set a meta tag to declare content-type UTF-8.

With this change, we insert this meta tag also when processing the request without using the API for translation. Any existing content-type declaration will be replaced.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
@@ -35,6 +35,7 @@ class HtmlConverter {
         removeHrefLangIfConflicts();
         appendSnippet(lang, type);
         appendHrefLang(headers);
+        replaceContentType();
         return doc.html();
     }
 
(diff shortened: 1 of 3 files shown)


a-bougouin:
On html-swapper, encoding is explicitly changed to UTF-8 all the time at parser level, so it makes sens to add the meta for UTF-8 encoding. Here I can't see UFT-8 encoding being set for parser output, so I am worried we add UTF-8 meta while HTML content not necessarily UTF-8. Do you have explanation?

torkelbe:
@a-bougouin The encoding change happens in `Utf8.java`, and is called from `WovnHttpServletResponse.java` like this
```
    public String toString() {
        return Utf8.toStringUtf8(this.getData());
    }
```
It is used when getting the original response from the server (in `WovnServletFilter.java`) like this
```
        String originalBody = wovnResponse.toString();
```

Whenever the request is processed by wovnjava (and the content is text, not binary), the output will be UTF-8. We also already set character encoding to UTF-8 in the HTTP header.

torkelbe:
@bstewart00 wovnjava currently supports the following content encodings from the servlet
* UTF-8
* Shift_JIS
* windows-31j
* EUC-JP
* x-euc-jp-linux
* x-eucJP-Open

Regardless of the encoding received, wovnjava will always output content as UTF-8.

bstewart00:
@torkelbe I wonder if that actually works in practice though, whether html swapper will decode the text correctly

torkelbe:
@bstewart00 I can confirm it works correctly for Shift_JIS and windows-31j, at the very least.

PR #91, opened 2019-04-09
Read project version dynamically

torkelbe:
We currently specify the wovnjava version in both `pom.xml` and in `Settings.java`.

With this change wovnjava will be able to read which version it was compiled as, as specified in `pom.xml`.


src/main/java/com/github/wovnio/wovnjava/Version.java
```diff
@@ -0,0 +1,23 @@
+package com.github.wovnio.wovnjava;
+
+import java.io.IOException;
+import java.util.Properties;
+
+class Version {
+    private static final String fallbackVersion = "projectPropertiesErrorFallback";
```
a-bougouin: I would set version to `"unknown"` instead.

torkelbe: I'm ok with that

PR #94, opened 2019-05-30
HTTP header debugging fields

torkelbe:
Improve debugging by including information about the request handling in HTTP header fields.

### Changes
* Include Fastly HTTP headers from the API response
    - `X-Wovn-Cache`
    - `X-Wovn-Cache-Hits`
    - `X-Wovn-Surrogate-Key`
* Set HTTP headers for the status of API request
    - `X-Wovn-Api`
    - `X-Wovn-Api-Status`
* Whenever wovnjava touches the request, declare itself in HTTP header
    - `X-Wovn-Handler`
* Adjusted ApiException management to make it simpler to declare such errors in `X-Wovn-Api` header field
* Remove dynamic setting of `data-wovnio-type` snippet attribute
    - The value is now always set to `fallback` for when wovnjava processes the HTML response
    - Once the widget has been updated, this attribute will declare wovnjava version
* In the cache key for translation API request, wovnjava will now include its version.
    - This make sure that behavior gets updated when the customer updates wovnjava

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Api.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Api.java
@@ -93,7 +93,7 @@ class Api {
                 }
                 return extractHtml(input);
             } else {
-                throw new ApiException("status_" + String.valueOf(status), "");
+                throw new ApiException("Failure", "Status code " + String.valueOf(status));
             }
         } catch (UnsupportedEncodingException e) {
             throw new ApiException("UnsupportedEncodingException", e.getMessage());
@@ -136,7 +136,7 @@ class Api {
         LinkedHashMap<String, String> dict = JSON.decode(json);
         String html = dict.get("body");
         if (html == null) {
-            throw new ApiException("unknown_json_format", "");
+            throw new ApiException("ResponseFormatError", "Unknown JSON format");
         }
         return html;
     }
--- a/src/main/java/com/github/wovnio/wovnjava/Api.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Api.java
@@ -13,41 +13,28 @@ import java.net.SocketTimeoutException;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 import java.util.LinkedHashMap;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Collections;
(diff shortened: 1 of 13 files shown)


src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -9,28 +9,45 @@
...
 import net.arnx.jsonic.JSON;
 
 class Api {
+    private static final Map<String, String> wovnFastlyHeaders;
+    static {
+        Map<String, String> headers = new HashMap<String, String>();
+        headers.put("X-Cache", "X-Wovn-Cache");
+        headers.put("X-Cache-Hits", "X-Wovn-Cache-Hits");
+        headers.put("X-Wovn-Surrogate-Key", "X-Wovn-Surrogate-Key");
+        wovnFastlyHeaders = Collections.unmodifiableMap(headers);
```
a-bougouin: This logic is outside of `Api` concern, I would make `HeadersForwarder` class of it.

torkelbe: Extracted methods for setting response headers to a separate class.

a-bougouin: thanks

bstewart00:
X-Wovn-Api is the one that feels a bit weird to me. What do we _mean_ by Api? I forgot what we intended to use this for.

torkelbe:
@bstewart00 `X-Wovn-Api` just asserts the status of the translation API request. We could change the HTTP header specification if we want the naming to be more specific.

src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -23,14 +24,17 @@
...
 
-    Api(Settings settings, Headers headers) {
+    Api(Settings settings, Headers headers, ResponseHeaders responseHeaders) {
         this.settings = settings;
         this.headers = headers;
+        this.responseHeaders = responseHeaders;
     }
 
     String translate(String lang, String html) throws ApiException {
+        this.responseHeaders.setApi("Requested");
```
bstewart00: if you feel like it, i'd rename the setters to `setApiStatus`, `setApiStatusCode`

PR #96, opened 2019-06-03
Add support for debugging query parameters

torkelbe:
### Changes
* Add server-side debugMode setting
* Query parameters to change the behavior of wovnjava
  - `wovnDisable`
    - always available
    - wovnajva does nothing
  - `wovnCacheDisable`
    - only available if debugMode is activated server-side
    - bypass translation API cache
  - `wovnDebugMode`
    - only available if debugMode is activated server-side
    - bypass translation API cache
    - send `debug_mode=true` to translation API
    - opportunity to embed debugging information in the response HTML
* Always send product and version information to translation API
* Bump version to 0.5.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ urlPattern                | yes      | 'path'
 query                     |          | ''
 defaultLang               | yes      | 'en'
 useProxy                  |          | 'false'
-debugMode                 |          | '0'
+debugMode                 |          | 'false'
 originalUrlHeader         |          | ''
 originalQueryStringHeader |          | ''
 ignoreClasses             |          | ''
@@ -147,7 +147,31 @@ When using a reverse proxy, if the WOVN.io Java Library is not given an appropri
 
 ### 2.7. debugMode
 
-As a development feature, if debugMode is set to 1, wovnjava will output debug logs.
+By turning on debugMode, you will enable extra debugging features for wovnjava.
+
+Turn on debugMode like this:
+```XML
+  <init-param>
+    <param-name>debugMode</param-name>
+    <param-value>true</param-value>
+  </init-param>
+```
+
+Two extra query parameters become available to change wovnjava's behavior:
+
+#### wovnCacheDisable
+Example request: `http://example.com/page/top.html?wovnCacheDisable`
+
+Using `wovnCacheDisable` as a query parameter will make wovnjava bypass the translation API cache, such that translation is always re-processed.
+This will make the request slower, but it is sometimes useful in order to force updated behavior.
+
+#### wovnDebugMode
+Example request: `http://example.com/page/top.html?wovnDebugMode`
+
+Using `wovnDebugMode` as a query parameter will activate embedded debug information in the response HTML comming from the server.
+This is intended to better understand what the problem is if something is not working correctly with wovnjava on your server.
+
+_Note that `wovnCacheDisable` and `wovnDebugMode` is only available when debugMode is turned on in your wovnjava configuration._
 
 ### 2.8. originalUrlHeader, originalQueryStringHeader
 
--- a/src/main/java/com/github/wovnio/wovnjava/Api.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Api.java
@@ -135,6 +135,8 @@ class Api {
         appendKeyValue(sb, "&lang_code=", lang);
         appendKeyValue(sb, "&url_pattern=", settings.urlPattern);
         appendKeyValue(sb, "&site_prefix_path=", settings.sitePrefixPathWithoutSlash);
+        appendKeyValue(sb, "&product=", "wovnjava");
+        appendKeyValue(sb, "&version=", Settings.VERSION);
         appendKeyValue(sb, "&body=", body);
         return sb.toString();
     }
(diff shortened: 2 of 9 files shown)


src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -155,6 +158,10 @@ private URL getApiUrl(String lang, String body) throws UnsupportedEncodingExcept
         appendValue(sb, lang);
         appendValue(sb, "&version=wovnjava_");
         appendValue(sb, Settings.VERSION);
+        if (headers.getCacheDisableMode() || headers.getDebugMode()) {
```
bstewart00: I don't think you necessarily want to do this for only debug mode

torkelbe: I believe we talked about only allowing cache bypass if debugMode was turned on server-side

bstewart00:
I would suggest moving as much as possible out of Headers. You may want to consider a RequestOptions class that takes in the request and exposes the various query parameters we have. I don't really like Headers holding state for them.

src/main/java/com/github/wovnio/wovnjava/Api.java
bstewart00: that is built-in in the `getCacheDisableMode` check isn't it? But just enabling debug mode you might want to use the cached version until you decide to override it.

README.md
```diff
@@ -147,7 +147,31 @@ When using a reverse proxy, if the WOVN.io Java Library is not given an appropri
...
+#### wovnCacheDisable
+Example request: `http://example.com/page/top.html?wovnCacheDisable`
+
+Using `wovnCacheDisable` as a query parameter will make wovnjava bypass the translation API cache, such that translation is always re-processed.
+This will make the request slower, but it is sometimes useful in order to force updated behavior.
+
+#### wovnDebugMode
+Example request: `http://example.com/page/top.html?wovnDebugMode`
+
+Using `wovnDebugMode` as a query parameter will activate embedded debug information in the response HTML comming from the server.
```
bstewart00: I don't think we need both a configuration setting + request parameter for debug mode. Just the setting should be enough?

src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -155,6 +160,10 @@ private URL getApiUrl(String lang, String body) throws UnsupportedEncodingExcept
         appendValue(sb, lang);
         appendValue(sb, "&version=wovnjava_");
         appendValue(sb, Settings.VERSION);
+        if (this.requestOptions.getCacheDisableMode() || this.requestOptions.getDebugMode()) {
+            appendValue(sb, "&timestamp=");
+            appendValue(sb, String.valueOf(System.currentTimeMillis()));
```
a-bougouin: I think you should use `appendKeyValue()` instead of the two lines.
If not, looking at `appendKeyValue()`, it seems `appendValue(sb, "&timestamp=");` should be replaced by `append(sb, "&timestamp=");`?

src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
```diff
@@ -86,4 +92,12 @@ private void tryTranslate(Headers headers, HttpServletRequest request, HttpServl
             out.close();
         }
     }
+
+    /*
+     * Check for `wovnDisable` in the query string
+     */
+    private boolean disableFilter(ServletRequest request) {
```
bstewart00: Can be moved to RequestOptions

src/main/java/com/github/wovnio/wovnjava/Api.java
torkelbe: `appendValue` is used because it encodes the appended string, and I believe the cache key needs to be fully encoded, as done here.

`appendKeyValue` is used in other places to encode only the value, not the key.

a-bougouin: I see, we probably can do better method naming (not here tho).

src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
```diff
@@ -30,6 +30,11 @@ public void init(FilterConfig config) throws ServletException {
     @Override
     public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws ServletException, IOException
     {
+        if (RequestOptions.wovnDisableMode((HttpServletRequest)request)) {
```
bstewart00: Not particularly a fan of making this static, even if it doesn't happen to use settings for that method. You construct an instance of RequestOptions further down in this method, so you can just move that up..

torkelbe: This checks happens in the main `doFilter` method, while I requestOptions is created and used in `tryTranslate` after deciding to process the request. I figured I wanted to do as little as possible before checking if we want to touch the request at all.

But I suppose I could instantiate it here and pass it down. It's what we do with `headers`, after all.

torkelbe: done

PR #98, opened 2019-06-06
Extract and test URL manipulation functions

torkelbe:
I want to clean up URL management in order to address some bugs related to it.

As part of that effort, I want to extract some URL manipulation functions into its own file so they can be tested separately. It will also let me use the functions outside of Headers if I need to.

The functions are moved over untouched.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/UrlPath.java
+++ b/src/main/java/com/github/wovnio/wovnjava/UrlPath.java
@@ -1,0 +1,49 @@
+package com.github.wovnio.wovnjava;
+
+final class UrlPath {
+    private UrlPath() {}
+
+    public static String removeFile(String path) {
+        if (path.endsWith("/")) {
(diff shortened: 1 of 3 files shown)


torkelbe:
Rebased to updated master

src/main/java/com/github/wovnio/wovnjava/UrlPath.java
```diff
@@ -0,0 +1,49 @@
...
+            return left + right.substring(1);
+        } else if (l || r) {
+            return left + right;
+        } else {
+            return left + "/" + right;
+        }
+    }
+
+    public static String normalize(String path) {
+        path = replaceRepeat(path, "/./", "/");            // remove redundant relative path
```
bstewart00: Probably would look to replacing these functions with the Java URI class at some point. Fairly sure this relative path code isn't quite right

PR #102, opened 2019-06-17
Simplify sitePrefixPath settings

torkelbe:
Site prefix path has two settings variables.
* `sitePrefixPathWithoutSlash` is the common format
* `sitePrefixPathWithSlash` is only being used one place

With these changes,
1. remove `sitePrefixPathWithSlash` and always access the common format
2. rename `sitePrefixPathWithoutSlash` to just `sitePrefixPath`

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Settings.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Settings.java
@@ -61,8 +61,8 @@ class Settings {
         }
 
         p = config.getInitParameter("sitePrefixPath");
-        this.hasSitePrefixPath = p != null && p.length() > 0;
-        if (this.hasSitePrefixPath) {
+        if (p != null && p.length() > 0) {
+            this.hasSitePrefixPath = true;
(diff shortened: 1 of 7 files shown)


src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -273,7 +270,7 @@ String hash() throws NoSuchAlgorithmException {
         for (String q : query) {
             md.update(q.getBytes());
         }
-        md.update(sitePrefixPathWithSlash.getBytes());
```
Soichiro725: I do not know much about this code.
However, I think that md will be changed if slash is removed.
Is this OK?

torkelbe: Good point. Yes, `md` will be slightly different here. It is used to create the settings hash that is part of the cache key for translation API requests. This should be ok because the cache key will be different for a new version of wovnjava anyways (the version number is included in the cache key).

PR #104, opened 2019-06-21
[Refactor] Create PatternHandlers

torkelbe:
This is the first of several PRs related to URL management refactoring that I will all merge into `refactor/url_management` before releasing.

I create three UrlLanguagePatternHandlers for doing URL manipulation

- PathUrlLanguagePatternHandler
- QueryUrlLanguagePatternHandler
- SubdomainUrlLanguagePatternHandler

The proper UrlLanguagePatternHandler is initialized at filter startup.

UrlLanguagePatternHandler should be responsible for

    getLang(url)
    removeLang(url)
    insertLang(url, lang)

In this PR, only the `getLang` operation is migrated to the UrlLanguagePatternHandlers. Its behavior from before remains unchanged. The other UrlLanguagePatternHandler functions are placeholders for now.

Many lines in test files are changed, but this is mostly because (1) initialization and Header constructor has changed, and (2) initialization may now throw ConfigurationError. The other change affecting tests is the removal of UrlPatternReg setting.

### Changes
* Create UrlLanguagePatternHandlers for managing URLs of urlPattern path, query, and subdomain
* Remove UrlPatternReg setting
* Hide regular expressions for URL management inside the UrlLanguagePatternHandlers
* Raise useful ConfigurationError at startup if library settings are invalid


src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -20,10 +20,12 @@
 
     private HttpServletRequest request;
     private String pathLang;
+    private PatternHandler patternHandler;
 
-    Headers(HttpServletRequest r, Settings s) {
+    Headers(HttpServletRequest r, Settings s, PatternHandler patternHandler) {
```
a-bougouin: For clarity I would add information to the name of the type. I would prefer `PatternHandler` to be call `UrlLangPatternHandler`.

src/main/java/com/github/wovnio/wovnjava/PatternHandler.java
```diff
@@ -0,0 +1,33 @@
+package com.github.wovnio.wovnjava;
+
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
+
+abstract class PatternHandler {
+    protected Pattern getLangPattern;
```
a-bougouin: I think `langRegex` is enough and will avoid the weirdness of `getGetLangPattern`. But since is it likely temporary, I guess it is not a huge deal.

src/main/java/com/github/wovnio/wovnjava/QueryPatternHandler.java
```diff
@@ -0,0 +1,23 @@
+package com.github.wovnio.wovnjava;
+
+import java.util.regex.Pattern;
+
+class QueryPatternHandler extends PatternHandler {
+    static final String QueryPatternRegex = "(?:(?:\\?.*&)|\\?)wovn=([^&]+)(?:&|$)";
```
a-bougouin: You might want to review naming convention either here or at https://github.com/WOVNio/wovnjava/pull/104/files#diff-a28a80824ff1f2d1b9b046bfe472ad20R6

src/main/java/com/github/wovnio/wovnjava/QueryPatternHandler.java
```diff
@@ -0,0 +1,23 @@
...
+    QueryPatternHandler() {
+        this.getLangPattern = Pattern.compile(QueryPatternRegex);
+    }
+
+    String getLang(String url) {
+        return this.getLangMatch(url, this.getLangPattern);
+    }
+
+    String removeLang(String url) {
+        return "site.com/path";
```
a-bougouin: PR is WIP?

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -238,7 +228,7 @@ public String locationWithLangCode(String location) {
         if (settings.urlPattern.equals("query") && location.contains("wovn=")) {
             return location;
         } else if (settings.urlPattern.equals("path")) {
-            Pattern p = Pattern.compile(settings.urlPatternReg);
+            Pattern p = this.patternHandler.getGetLangPattern();
```
bstewart00: Seems like you could replace this with patternHandler.getLang.

src/main/java/com/github/wovnio/wovnjava/QueryPatternHandler.java
torkelbe: As the PR description indicates, this is the first of several PRs related to URL management refactoring that I will all merge into `refactor/url_management` before releasing.

In this PR, only the `getLang` operation is migrated to the PatternHandlers. Its behavior from before remains unchanged. The other PatternHandler functions are placeholders for now.

src/main/java/com/github/wovnio/wovnjava/PatternHandler.java
torkelbe: It was originally called `langPattern`, but afterwards patterns for remove lang and insert lang were needed, so I renamed it. I agree, `getGet...` is not great. It should be temporary, however.

src/main/java/com/github/wovnio/wovnjava/Headers.java
torkelbe: You're right, PatternHandler` is not great. How about `UrlPatternHandler` to mirror the `urlPattern` setting?

src/main/java/com/github/wovnio/wovnjava/Headers.java
torkelbe: I think you're right. I hesitated to change this function for now, but I agree it's better to use `getLang` directly. That also lets me hide `getLangPattern`.

src/main/java/com/github/wovnio/wovnjava/PatternHandler.java
torkelbe: I have removed outside access to `getLangPattern`, so there is no longer need for the wonky getter.

src/main/java/com/github/wovnio/wovnjava/QueryPatternHandler.java
torkelbe: Renamed consistently now.

src/main/java/com/github/wovnio/wovnjava/Headers.java
a-bougouin: I never thought `urlPattern` was good enough :'(
I think "language" is an important information that should be in the name.

src/main/java/com/github/wovnio/wovnjava/PatternHandler.java
a-bougouin: :+1:

src/main/java/com/github/wovnio/wovnjava/QueryPatternHandler.java
a-bougouin: thanks

src/main/java/com/github/wovnio/wovnjava/QueryPatternHandler.java
a-bougouin: I saw that afterwards. Test would be good for implemented stuffs tho.

src/main/java/com/github/wovnio/wovnjava/Headers.java
torkelbe: Renamed PatternHandler to UrlLangPatternHandler

bstewart00: May as well go ahead with UrlLanguagePatternHandler. Not a huge fan of abbreviations in class names.

a-bougouin: Thanks

torkelbe:
@a-bougouin Regarding unit test, I would like to defer that to when we change the behavior. While these new classes definitely should have tests, their behavior is occasionally broken in isolation. I made it a point to not change the logic of the code for now, and to avoid any test regression. Rest assured, the process for reading language code from the incoming request will change soon.

a-bougouin:
I forgot about the tests. I won't at https://github.com/WOVNio/wovnjava/pull/105

PR #105, opened 2019-06-21
[Refactor] Move removeLang into PatternHandlers

torkelbe:
Move `removeLang` logic from Headers into the respective PatternHandlers.

Logic remains unchanged.


torkelbe:
Rebased to include previous changes to PatternHandlers

PR #106, opened 2019-06-21
[Refactor] Calculate requestLang at Headers initialization

torkelbe:
### Changes
* Rename `pathLang` to `requestLang`.
* Calculate `requestLang` only once, when Headers is instantiated. Change the getter to simply return the value.

Logic for calculating the value remains unchanged.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -27,7 +27,7 @@ class Headers {
         this.request = r;
         this.urlLanguagePatternHandler = urlLanguagePatternHandler;
 
-        this.requestLang = this.calculateRequestLang();
+        this.requestLang = this.computeRequestLang();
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -268,12 +258,25 @@ String getUrlWithoutLanguageCode() {
...
+            lang = this.requestLang;
         }
         return this.patternHandler.removeLang(uri, lang);
     }
 
     boolean isValidPath() {
         return this.pathName.startsWith(this.settings.sitePrefixPath);
     }
+
+    private String calculateRequestLang() {
```
a-bougouin: I would use "compute" over "calculate", but that's nitpicking...

torkelbe:
I rebased these changes to be updated with the modified PatternHandler refactor.

The changes remain the same, with the additional change of using "compute" over "calculate".

PR #108, opened 2019-06-24
Redesign URL management

torkelbe:
Partially redesign URL management, and correctly handle client request URLs.

### Overall changes
* (Refactor) Create UrlLanguagePatternHandlers for managing language code in URLs
* Match languages only by their language code
* Compute client request URL based on HTTP headers and servlet request attributes
* Process each request only once

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Lang.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Lang.java
@@ -1,94 +1,58 @@
 package com.github.wovnio.wovnjava;
 
-import org.jetbrains.annotations.Contract;
-
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
 class Lang {
-    static final Map<String, Lang> LANG;
+    private static final Map<String, Lang> LANG;
     static {
         HashMap<String, Lang> map = new HashMap<String, Lang>();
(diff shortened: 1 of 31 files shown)


src/main/java/com/github/wovnio/wovnjava/Lang.java
```diff
@@ -1,94 +1,58 @@
...
-            String langNameLC = langName.toLowerCase();
-            if ( langNameLC.equals(l.name.toLowerCase())
-                    || langNameLC.equals(l.en.toLowerCase())
-                    || langNameLC.equals(l.code.toLowerCase())
-                    ) {
-                return l.code;
-            }
-        }
-        return null;
+    Lang(String code, String hreflangCode, String englishName, String nativeName) {
```
a-bougouin: I would like not to loose the information that `hreflangCode` is actually the ISO639_1 code, so I would prefer you to rename `hreflangCode`, please.

torkelbe: Adressed in https://github.com/WOVNio/wovnjava/pull/118

a-bougouin: thanks

PR #109, opened 2019-07-01
Allow Lang lookup only by language code

torkelbe:
Change to allow Lang lookup only by language code. This fixes some incorrect behaviors in URL handling. `Lang.java` is greatly simplified in the process.

Note that in the Lang initialization, information for the names of languages is kept in the file, but it is no longer being used by the library.

### Changes
* Lang object lookup only uses language code now
    - This affects URL matches and settings configuration
* Raise ConfigurationError if the setting for defaultLang or supportedLangs is invalid
    - This is added now because a configuration like `defaultLang=English` becomes invalid
* Access ISO639-1 language code for hreflang using a Lang field instead of conversion function
* (Whitespace changes) retab test files for UrlLanguagePatternHandlers

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Settings.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Settings.java
@@ -187,7 +187,7 @@ class Settings {
     }
 
     private void initialize() {
-        this.defaultLang = Lang.getCode(this.defaultLang);
+        this.defaultLang = Lang.getLang(this.defaultLang).code;
 
         if (this.supportedLangs.size() == 0) {
             this.supportedLangs.add(this.defaultLang);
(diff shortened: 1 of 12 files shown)


src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -186,8 +186,16 @@ static boolean getBoolParameter(String param) {
...
     }
 
-    private void initialize() {
-        this.defaultLang = Lang.getCode(this.defaultLang);
+    private void initialize() throws ConfigurationError {
+        if (Lang.getLang(this.defaultLang) == null) {
+            throw new ConfigurationError("Invalid language code for defaultLang: " + this.defaultLang);
+        }
+
+        for (int i = 0; i < this.supportedLangs.size(); i++) {
```
bstewart00: I'd use a foreach loop here: `for (String supportedLang : this.supportedLangs)`

PR #110, opened 2019-07-02
Fix and improve tests for removeLang

torkelbe:
Fix bugs with `removeLang` for when `urlPattern=path`, and improve test cases for all removeLang variations.

### Changes
* Fix PathUrlLanguagePatternHandler to only match path lang code at the beginning of the path
* Add test cases for when the input URL includes schema (this was also broken for urlPattern path)
    - now we should be able to pass a URL of any form into `removeLang` and see expected behavior

### DIFF
--- a/src/test/java/com/github/wovnio/wovnjava/QueryUrlLanguagePatternHandlerTest.java
+++ b/src/test/java/com/github/wovnio/wovnjava/QueryUrlLanguagePatternHandlerTest.java
@@ -30,13 +30,14 @@ public class QueryUrlLanguagePatternHandlerTest extends TestCase {
         assertEquals("/page/index.html", sut.removeLang("/page/index.html", "ja"));
         assertEquals("/page/?wovn=en", sut.removeLang("/page/?wovn=en", "ja"));
         assertEquals("ja.site.com/ja/?lang=ja", sut.removeLang("ja.site.com/ja/?lang=ja", "ja"));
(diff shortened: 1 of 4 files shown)


src/test/java/com/github/wovnio/wovnjava/QueryUrlLanguagePatternHandlerTest.java
```diff
@@ -28,15 +28,17 @@ public void testRemoveLang__NonMatchingQuery__DoNotModify() {
...
         assertEquals("/page/?wovn=en", sut.removeLang("/page/?wovn=en", "ja"));
         assertEquals("ja.site.com/ja/?lang=ja", sut.removeLang("ja.site.com/ja/?lang=ja", "ja"));
+        assertEquals("http://site.com/page/?wovn=japan", sut.removeLang("http://site.com/page/?wovn=japan", "ja"));
     }
 
     public void testRemoveLang__MatchingQuery__RemoveLangCode() {
         QueryUrlLanguagePatternHandler sut = new QueryUrlLanguagePatternHandler();
         assertEquals("", sut.removeLang("?wovn=ja", "ja"));
         assertEquals("/?search=pizza&lang=ja", sut.removeLang("/?search=pizza&wovn=ja&lang=ja", "ja"));
-        assertEquals("site.com/page/index.html?wovn&wovn", sut.removeLang("site.com/page/index.html?wovn&wovn=ja&wovn", "ja"));
```
torkelbe: Adjusted this test case because no test did the form `path/index.html?some=query?wovn=en` with the lang query at the end.

PR #112, opened 2019-07-06
Compute client request URL

torkelbe:
Create static class UrlResolver for correctly computing client request URL. The priority for URL fields to use for this purpose is
1. HTTP headers
2. Internal servlet container request forwarding attributes
3. Current request values

Using values from HTTP headers is dependent on user configuration (such as `useProxy` setting). If configuration is set to use HTTP headers but the relevant HTTP headers are not set, UrlResolver will resolve the URL normally.

Current requestLang is determined based on the client request URL.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -21,16 +21,14 @@ class Headers {
     private HttpServletRequest request;
     private String requestLang;
     private UrlLanguagePatternHandler urlLanguagePatternHandler;
-    private String clientRequestUrl;
 
     Headers(HttpServletRequest r, Settings s, UrlLanguagePatternHandler urlLanguagePatternHandler) {
         this.settings = s;
         this.request = r;
         this.urlLanguagePatternHandler = urlLanguagePatternHandler;
 
-        this.requestLang = this.computeRequestLang();
-
-        this.clientRequestUrl = UrlResolver.computeClientRequestUrl(r, s);
+        String clientRequestUrl = UrlResolver.computeClientRequestUrl(r, s);
+        this.requestLang = this.urlLanguagePatternHandler.getLang(clientRequestUrl);
 
         this.protocol = this.request.getScheme();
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -27,7 +27,8 @@
         this.request = r;
         this.urlLanguagePatternHandler = urlLanguagePatternHandler;
 
-        this.requestLang = this.computeRequestLang();
+        String clientRequestUrl = UrlResolver.computeClientRequestUrl(r, s);
```
bstewart00: definitely rename r and s parameters

bstewart00: Not a huge fan of making it static though, it's the kind of thing I think should be passed in as a parameter to Headers, `new UrlResolver(request, settings)`

src/main/java/com/github/wovnio/wovnjava/UrlResolver.java
```diff
@@ -0,0 +1,92 @@
...
+        String path = clientRequestPath(request, settings.originalUrlHeader);
+        String query = clientRequestQuery(request, settings.originalQueryStringHeader);
+        return scheme + "://" + host + path + query;
+    }
+
+    private static String clientRequestHostAndPort(HttpServletRequest request, boolean useProxy) {
+        String host = null;
+        Integer port = null;
+        if (useProxy) {
+            /* request.getHeader returns String or null */
```
a-bougouin: minor: I would use single line comment

src/main/java/com/github/wovnio/wovnjava/Headers.java
torkelbe: I will ask to keep UrlResolver as a static class for now. The reason is that it keeps this set of changes short and concise.

In essence, we currently both compute request URL and find its request language inside `Headers.java`. Now, the "compute request URL" part is moved outside and tested separately, but otherwise `Headers.java` has the same purpose and design as before.

`Headers.java` will still receive a lot of changes, and how to best combine the components will become easier to see as we clean up the old code.

PR #113, opened 2019-07-10
[Refactor] Set request state final variables

torkelbe:
### Changes
* Set final variables for request state in Headers constructor
    - `clientRequestUrlWithoutLangCode`
    - `shouldRedirectToDefaultLang`
    - `isValidPath`
* Use UrlLanguagePatternHandler to check if request URL matches sitePrefixPath
* Ensure that `removeLang` accepts being called with empty language argument

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Interceptor.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Interceptor.java
@@ -16,7 +16,7 @@ class Interceptor {
     }
 
     String translate(String body) {
-        String lang = headers.requestLang;
+        String lang = headers.getRequestLang();
         boolean canTranslate = lang.length() > 0 && !lang.equals(settings.defaultLang);
         if (canTranslate) {
             return apiTranslate(lang, body);
--- a/src/main/java/com/github/wovnio/wovnjava/Interceptor.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Interceptor.java
@@ -16,7 +16,7 @@ class Interceptor {
     }
 
     String translate(String body) {
-        String lang = headers.getRequestLang();
+        String lang = headers.requestLang;
         boolean canTranslate = lang.length() > 0 && !lang.equals(settings.defaultLang);
         if (canTranslate) {
             return apiTranslate(lang, body);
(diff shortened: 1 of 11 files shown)


src/main/java/com/github/wovnio/wovnjava/Interceptor.java
```diff
@@ -16,7 +16,7 @@
     }
 
     String translate(String body) {
-        String lang = headers.getRequestLang();
+        String lang = headers.requestLang;
```
a-bougouin: I am not a huge fan of having `requestLang` public. It breaks uniform access, now we use either public attributes or getters. I prefer using getters for consistency.

src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
```diff
@@ -26,6 +30,10 @@ String insertLang(String url, String lang) {
         return "site.com/en/path";
     }
 
+    boolean isMatchSitePrefixPath(String url) {
```
a-bougouin: I think it should be named `isMatchingSitePrefixPath` or `matchesSitePrefixPath`.

src/main/java/com/github/wovnio/wovnjava/UrlLanguagePatternHandler.java
```diff
@@ -10,6 +10,8 @@
 
     abstract String insertLang(String url, String lang);
 
+    abstract boolean isMatchSitePrefixPath(String url);
```
a-bougouin: I would not make it abstract and implement default returning `true`. That would avoid having to think about it when implementing URL handler for non-path URL components. Only concerned handlers would override it.

Both solution are fine. I prefer the one I mentioned so I wanted to bring that up.

torkelbe: That's nicer

src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
torkelbe: Now called `isMatchingSitePrefixPath`

src/main/java/com/github/wovnio/wovnjava/Interceptor.java
torkelbe: Headers currently has almost no getters, and fields are simply accessed directly. Consistency points to not using getters, if anything. That said, the old fields that are accessed directly will likely be removed soon.

I think I liked accessing the fields as if Headers is a struct, essentially. But I'm fine with using getters as well.

PR #114, opened 2019-07-11
[Misc] Remove references to HTTP header "Location"

torkelbe:
HTTP Header `Location` is currently not used, but it is being mocked and "tested"


torkelbe:
Wovnjava handles redirections, just not by manipulating the `Location` HTTP header directly.

PR #115, opened 2019-07-14
Process each request only once

torkelbe:
WovnServletFilter may intercept a single request multiple times if the request is forwarded internally within the servlet container. For instance, this forwarding happens when WovnServletFilter forwards a request with language code to the same request without language code.

WovnServletFilter will always set HTTP header `X-Wovn-Handler` when intercepting a request. Therefore, with these changes we will check if that header is set on the response object before processing the request. If the header is set, we will ignore it and pass it through untouched.

This change is more important now than before because we now calculate the request language etc based on the client requested URL, not the current servlet URL.


torkelbe:
@bstewart00 That sounds accurate.

The forwarding happens like this
```java
request.getRequestDispatcher("site.com/new/path").forward(request, response);
```
and just makes the servlet container re-process the request internally with the new URL (which will also trigger wovnjava again.)

PR #116, opened 2019-07-15
[Fix] Path pattern removeLang

torkelbe:
For path patterns, `removeLang` has incorrect interaction with query parameters in URL when path is empty (example: `site.com?query`). These changes apply a fix for it and expands test cases.

Also, the changes include a fix to trailing slash after removing language code. If the path ends like `/path/ja`, the resulting path without language code should not have trailing slash, like `/path`. If the original path ends with slash like `/path/ja/`, the resulting path should be `/path/`.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
+++ b/src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
@@ -23,7 +23,7 @@ class PathUrlLanguagePatternHandler extends UrlLanguagePatternHandler {
 
         Pattern removeLangPattern = buildRemoveLangPattern(lang);
         Matcher matcher = removeLangPattern.matcher(url);
-        return matcher.replaceFirst("$1$2$3");
+        return matcher.replaceFirst("$1$2$3$5");
     }
 
     String insertLang(String url, String lang) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
```diff
@@ -47,10 +47,10 @@ private Pattern buildGetLangPattern(String sitePrefixPath) {
...
         Pattern p = Pattern.compile(
                 "^(.*://)?" + /* optional schema */
-                "([^/]*)?" + /* optional host */
-                "(" + this.sitePrefixPath + "/)" + /* sitePrefixPath */
-                "(" + lang + ")" + /* lang code */
-                "(/|$)"
+                "([^/?]*)?" + /* optional host */
+                "(" + this.sitePrefixPath + ")" +
+                "(/" + lang + ")" +
+                "(/|\\?|$)" /* next path, query, or end-of-string */
```
Soichiro725: I think there is a possibility that `#` will continue.

torkelbe: Would be possible if we were working in the widget. But the server will not handle fragments, so fortunately we don't have to worry about it.

PR #117, opened 2019-07-15
Create hreflang URLs based on client request URL

torkelbe:
### Changes
* Implement `insertLang` for all UrlLanguagePatternHandlers
* Calculate hreflang target URLs in Headers based on the client request URL.
  - Use UrlLanguagePatternHandlers to insert language codes.
* Remove old `redirectLocation` as it is no longer used.
* Note that some hreflang test had the order of hreflang elements changed, but the contents of the elements are tested the same.


src/test/java/com/github/wovnio/wovnjava/HeadersTest.java
```diff
@@ -504,4 +436,59 @@ public void testOriginalHeaders() throws ConfigurationError {
         assertEquals("?baz=123", h.query);
         assertEquals("example.com/foo/bar?baz=123", h.pageUrl);
     }
+
```
a-bougouin: Some removed test cases removed above are not preserved. I think site prefix path test case should be preserved at least. I would also add test cases for custom language aliases as well as custom domain.

torkelbe: There is no need to preserve all test cases for the old redirectLocation in this file. It is no longer the responsibility of Headers.java to ensure that language codes are added correctly.

I guess site prefix path could be used for the `testGetHreflangUrlMap__PathPattern` test, but it does not make much difference for testing the correctness of the `getHreflangUrlMap` function's responsibilities.

wovnjava does not have custom language aliases or custom domain.

a-bougouin: I though some of the logic still would require test here.
But I review my comment. I agree it is not necessary :+1:

PR #119, opened 2019-07-18
[Fix] Send correct client request URL in translation API request

torkelbe:
Update translation API request to use the computed client request URL to identify which page is being translated.

The API request puts this information in two places. It sends the full URL in the API body, and it sends just the path in the cache key. We therefore have to extract the path part, and a new helper function is added for this purpose.

We use quite a bit of regex manipulation for URLs at this point. A lot of it is unavoidable, of course, and I'm trying to think about how to keep it to a minimum. That said, as long as any regex manipulation is tested in isolation, it feels manageable.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -24,7 +24,6 @@ class Headers {
 
     private final String requestLang;
     private final String clientRequestUrlWithoutLangCode;
-    private final String clientRequestPathWithoutLangCode;
     private final boolean shouldRedirectToDefaultLang;
     private final boolean isValidPath;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -24,6 +24,7 @@
 
     private final String requestLang;
     private final String clientRequestUrlWithoutLangCode;
+    private final String clientRequestPathWithoutLangCode;
```
Soichiro725: I think it is better to calculate with function instead of defining new member variable.
I want to reduce member variables as much as possible.

torkelbe: Hm, I have been moving in the opposite direction, to be honest.

It feels easier to understand and reason around the state of the request handling when we simply compute the various state variables up front, and never allow them to change.

So I don't think we want to reduce member variables as much as possible. However, I do think this particular variable a good candidate to compute as needed, in the getter. Also, the variable is only used once, at least for now.

I'll make the change, thanks.

PR #122, opened 2019-08-13
Remove query setting

torkelbe:
The purpose of the `query` setting is to identify unique pages when the specified query parameters are different. This is a setting that the customer configures for their Wovn project.

The standard behavior for our products is to include all query parameters in the request for page data, and the server will identify the page correctly. Therefore, wovnjava does not need to concern itself with the details of query parameter filtering, and we can remove the setting.

Existing configuration files do not need to be changed.

**Old behavior:** Send only the query parameters required to identify the page
**New behavior:** Send all query parameters always, and let the server pick query parameters for page identification

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -111,29 +111,8 @@ class Headers {
         }
         this.url += this.removeLang(this.query, this.langCode());
         this.url = this.url.length() == 0 ? "/" : this.url;
-        if (this.settings.query.size() > 0) {
-            ArrayList<String> queryVals = new ArrayList<String>();
-            for (String q : this.settings.query) {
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -111,29 +111,8 @@
...
-                for (String q : queryVals) {
-                    this.query += q + "&";
-                }
-                // remove last ampersand.
-                this.query = Pattern.compile("&$").matcher(this.query).replaceFirst("");
-            }
-        } else {
-            this.query = "";
+        if (!this.query.isEmpty() && !this.query.startsWith("?")) {
+            this.query = "?" + this.query;
```
torkelbe: Note that this modification is required to match the expected format that resulted from the old code. It is not elegant, but `this.query` should be considered a legacy part of `Headers.java`, and will be phased out soon.

PR #123, opened 2019-11-01
Refactor Settings

torkelbe:
Rewrite Settings class to have required and optional settings, sane default values, and useful error messages when incorrect configuration is used.

### Comments
The following configuration fields are now required
* `projectToken`
* `urlPattern`
* `defaultLang`
* `supportedLangs`

These fields no longer have default values, and wovnjava will throw an error if the servlet is started and these fields are not set. Therefore, if their current configuration is not complete, a user may have to fix their configuration after updating wovnjava with this change.

Note that various test initialization is updated to use Settings with valid required fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,6 @@ Parameter Name            | Required | Default Setting
 ------------------------- | -------- | ------------
 projectToken              | yes      | ''
 urlPattern                | yes      | 'path'
-query                     |          | ''
 defaultLang               | yes      | 'en'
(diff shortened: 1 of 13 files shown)


Soichiro725:
> These fields no longer have default values, and wovnjava will throw an error if the servlet is started and these fields are not set. Therefore, if their current configuration is not complete, a user may have to fix their configuration after updating wovnjava with this change.

We should notify this to es team.

PR #124, opened 2019-11-05
Refactor relative URL manipulation

torkelbe:
Refactor relative URL manipulation to take advantage of `java.net.URL` and our existing UrlLanguagePatternHandlers. This lets us remove a lot of the URL handling code inside Headers.

Wovnjava needs to parse URLs relative to the request location when the wrapped servlet context wants to redirect to a relative URL.

For example, if we're processing a request for `http://example.com/en/japan/index.html` and the servlet wants to redirect to `/global/index.html`, we must make sure that the redirect location keeps its language code like `http://example.com/en/global/index.html`.


src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -129,74 +139,34 @@ String langCode() {
         }
     }
 
+    /*
+     * Take as input a location string of any form (relative path, absolute path, absolute URL).
+     * If the location needs a Wovn language code, return an absolute URL string of that location
+     * with language code of the current request language. Else return the location as-is.
+     *
+     * Do not modify `location` if
```
bstewart00: Personally I think this code is pretty self explanatory and reads almost the same as the comment, so I would consider it redundant.

torkelbe: Fair enough 👍

src/main/java/com/github/wovnio/wovnjava/UrlContext.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public boolean isSameHost(URL url) {
+        return this.context.getHost().equalsIgnoreCase(url.getHost());
+    }
+
+    /*
+     * Remove sections of the URL path that starts with "/.."
+     */
+    private URL stripSlashDotDot(URL url) throws MalformedURLException {
```
Soichiro725: I want to add test case for location which start with `/..` and `/../..` .

torkelbe: I added a test case for absolute path starting with `/../..`.
I believe the test cases now cover relative path navigation, as well as correctly stripping multiple `/..` at the beginning of a path.

src/test/java/com/github/wovnio/wovnjava/HeadersTest.java
```diff
@@ -333,8 +333,8 @@ public void testLocationWithSitePrefixPath() throws ConfigurationError {
         assertEquals("https://example.com/global/th/", h.locationWithLangCode("https://example.com/global/th/"));
         assertEquals("https://example.com/global/ja/tokyo/", h.locationWithLangCode("https://example.com/global/tokyo/"));
         assertEquals("https://example.com/global/ja/file.html", h.locationWithLangCode("https://example.com/global/file.html"));
-        assertEquals("https://example.com/global/ja/file.html", h.locationWithLangCode("https://example.com/pics/../global/file.html"));
-        assertEquals("https://example.com/global/../../file.html", h.locationWithLangCode("https://example.com/global/../../file.html"));
+        assertEquals("https://example.com/pics/../global/file.html", h.locationWithLangCode("https://example.com/pics/../global/file.html"));
```
Soichiro725: Is this OK to change expected result?

torkelbe: This test (and the test below) asserts what happens if a servlet wants to redirect to a URL with explicit `/../` sections in its path.
```
language code to insert: "/ja/"
sitePrefixPath: "/global/"
URL to redirect to: "https://example.com/pics/../global/file.html"
```
I would like to say that the URL in question does not match their site prefix path because the string does not start with `"https://example.com/global/"`. If the URL was normalized it would match, but preferably it should be normalized before we receive it.

Of course, if they want to redirect to a relative path like `../global/file.html`, we handle the URL normalization properly. So the issue is only for redirects to absolute URLs that are not formatted correctly.

That said, I can take another look to see if we can conveniently normalize any URL string that we receive, including absolute ones.

torkelbe: I changed UrlContext to now also normalize non-normalized URL inputs. It makes our test behavior identical to before.

I originally wanted to not handle these cases because it requires us to do more processing on the input we receive, and that can easily have weird bugs or side effects. (It helps if we can make _some_ assumptions about our input.) I think this normalization method is ok, though.

PR #125, opened 2019-11-07
Create FileExtensionMatcher and clean up Headers

torkelbe:
This PR makes two structural changes. The primary purpose is to clean `Headers.java` of legacy URL parsing code.

### Clean up legacy code in `Headers.java`
The Headers class has several fields and a lot of code related to request URL that we are removing.
```java
    String host;
    String pathName;
    String pathNameKeepTrailingSlash;
    String protocol;
    String pageUrl;
    String query;
    String url;
```
Only two of these fields are still in use outside of Headers, `pathName` and `pathNameKeepTrailingSlash`. They are used for purposes that need the _location of the current servlet context_.

For all other request URL handling, we need the _location originally requested by the client_. This URL handling is already moved to `UrlResolver.java` (and is now stored in Headers as `clientRequestUrlWithoutLangCode`).

Instead of the old `pathName` and `pathNameKeepTrailingSlash`, we now store `currentRequestPathWithoutLangCode` that is simply computed from `HttpServletRequest.getRequestURI()`.

That means we can remove all the old URL processing code that still resides in Headers' constructor.

### Create `FileExtensionMatcher.java`
`currentRequestPathWithoutLangCode` described above is used for checking if the request is for a file type that we cannot translate. This regex processing is currently included in `HtmlChecker.java`.

This PR simply moves the file extension regex processing out of HtmlChecker and into a dedicated file, since it's a separate concern from checking HTML content.

Regex processing remains the same except for one adjustment to image file extension matching: `(?!jp$)jpe?g?` is changed to `jpe|jpe?g` to match the same adjustment made to our other products.

### DIFF
--- a/src/test/java/com/github/wovnio/wovnjava/FileExtensionMatcherTest.java
+++ b/src/test/java/com/github/wovnio/wovnjava/FileExtensionMatcherTest.java
@@ -12,7 +12,6 @@ public class FileExtensionMatcherTest extends TestCase {
     public void testIsFile() {
         assertIsFile(false, "");
         assertIsFile(false, "/");
-        assertIsFile(false, "/");
         assertIsFile(false, "html");
         assertIsFile(false, "jp");
         assertIsFile(true, "png");
--- a/src/test/java/com/github/wovnio/wovnjava/FileExtensionMatcherTest.java
+++ b/src/test/java/com/github/wovnio/wovnjava/FileExtensionMatcherTest.java
@@ -1,0 +1,53 @@
+package com.github.wovnio.wovnjava;
+
+import junit.framework.TestCase;
+
+public class FileExtensionMatcherTest extends TestCase {
+    private FileExtensionMatcher sut;
+
+    protected void setUp() {
+        this.sut = new FileExtensionMatcher();
+    }
+
+    public void testIsFile() {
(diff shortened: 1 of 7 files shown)


src/test/java/com/github/wovnio/wovnjava/FileExtensionMatcherTest.java
```diff
@@ -0,0 +1,53 @@
...
+    private FileExtensionMatcher sut;
+
+    protected void setUp() {
+        this.sut = new FileExtensionMatcher();
+    }
+
+    public void testIsFile() {
+        assertIsFile(false, "");
+        assertIsFile(false, "/");
+        assertIsFile(false, "/");
```
Soichiro725: There is same case.
Original test case had a `null` pattern.

torkelbe: > There is same case.
Thanks! Removed now

> Original test case had a null pattern.
Yes, I removed the `null` input handling from the function. It originally begins with
```java
        if (path == null) {	
            return true;	
        }
```
But the input (`HttpServletRequest.getRequestURI()`) cannot be `null`; the servlet context always has a path, even if it is empty string. So we should remove the null-case.

PR #127, opened 2019-11-08
Base UrlContext on the current servlet context URL

torkelbe:
### Internal forwarding
UrlContext is used as base URL for forwarding relative URLs internally.

We intercept the forwarding call because we want to make sure to strip Wovn language code from the current URL context, such that it is as transparent as possible to introduce wovnjava to an existing server.

Currently, UrlContext is based on the client request URL. However, if the request has been altered by a reverse proxy, the current URL context may be different from the original request, and our "base URL" may be incorrect at the point where the internal forwarding happens.

### Solution
I want to change the UrlContext object to be based on the location of the current internal context.

In summary, when manipulating a request URL that may have changed, I think we should
1) Determine the language of the request from the original URL sent by the client.
2) "Strip language" only from the current context URL, if possible.

_Note: We will not try to hide language code from the original URL sent by the client._

### Examples
**Example 1)** Successfully stripping language code from path

| url pattern | client request URL | URL received by wovnjava from reverse proxy |
|---|---|---|
| `path` | `http://site.com/en/cat.png` | `http://internal.site.com/en/cat.png` |

In this case, we determine from the client request URL that the language is `/en/`, and we can successfully strip that language code from the current context URL, such that the servlet below wovnjava will receive a request for the URL `http://internal.site.com/cat.png`.

**Example 2)** Not possible to strip language code from current context

| url pattern | client request URL | URL received by wovnjava from reverse proxy |
|---|---|---|
| `subdomain` | `http://en.site.com/cat.png` | `http://internal.site.com/cat.png` |

In this case, we determine from the client request URL that the language is `/en/`, but we _cannot_ successfully strip that language code from the current context URL. Therefore, we pass the URL as-is to the servlet below wovnjava, which will receive a request for the URL `http://internal.site.com/cat.png`.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -17,8 +17,6 @@ class Headers {
     private final String requestLang;
     /* The URL that the client originally requested */
     private final String clientRequestUrlInDefaultLanguage;
-    /* The path of the current servlet context */
-    private final String currentContextPathInDefaultLanguage;
 
     private final boolean shouldRedirectToDefaultLang;
     private final boolean isValidRequest;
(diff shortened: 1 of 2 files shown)


torkelbe:
@bstewart00 When wovnjava receives a request that has been rewritten, we already have solutions for inspecting the URL that was originally sent by the client. We use that to determine the language of the request.

If the request was rewritten within the java web server, built-in variables are available for inspecting the URL first received by the java web server.

If the request was rewritten by a web server in front of java (nginx etc), HTTP headers can be configured to allow wovnjava to see the original request URL.

bstewart00:
Sounds like everything is good then. Ideally a clear distinction can be made in the code between them.

PR #130, opened 2019-11-18
Always forward to path in default language

torkelbe:
Currently when we intercept a request, we sometimes forward it with a different path (internally in the servlet container). This happens when the request path is different from the path in default language, such as when receiving `site.com/en/page.html` and `urlPattern = path`. We will forward that request to `site.com/page.html`.

However when we _don't_ need to translate the incoming request, such as for an image resource, we currently never forward it. Instead, the wrapper for the request object will try to hide language codes when accessing information about the path.

Hiding language codes from the getters on the request object is difficult because of how the servlet context influences the correct response, particularly for getters on the URL path. This stackoverflow post illustrates the meaning of some of these getters https://stackoverflow.com/a/21046620.

### Solution
Whenever the request has a path that is different from the equivalent path in default language, we want to forward it to the default language. That way, the servlet container will create the correct request object, and we do not have to wrap the getters for path information.

We still want to wrap the information about hostname, so we keep the `@Override` functions for `getServerName` and `getRequestURL`.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Headers.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Headers.java
@@ -18,7 +18,11 @@ class Headers {
     /* The URL that the client originally requested */
     private final String clientRequestUrlInDefaultLanguage;
 
+    /* Should send HTTP 302 redirect to page in default language */
     private final boolean shouldRedirectToDefaultLang;
+    /* Should forward the request internally to equivalent path in default language */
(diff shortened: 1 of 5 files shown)


torkelbe:
@bstewart00 Thanks for the comment. Now that you mention it, I'm also not a big fan of these `shouldForwardToPathInDefaultLanguage` names. For this variable in particular, it feels like it unnecessarily hides reasons for why the path should be forwarded. I changed the name now, and it looks better.

PR #132, opened 2019-11-20
Ignore URL language codes that are not declared in supportedLangs

torkelbe:
When checking language code of a URL, ignore any language code that is not declared by the user in `supportedLangs` setting.

For example, if supported languages include only English and Japanese, do not consider the `/th/` in `http://site.com/th/page` as a Thai language code.

Also, refactor `getLang` to return Lang object instead of String, or null if no language code is found.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
+++ b/src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
@@ -1,14 +1,19 @@
 package com.github.wovnio.wovnjava;
 
+import java.util.ArrayList;
 import java.util.regex.Pattern;
 import java.util.regex.Matcher;
 
 class PathUrlLanguagePatternHandler extends UrlLanguagePatternHandler {
+    private Lang defaultLang;
+    private ArrayList<Lang> supportedLangs;
     private String sitePrefixPath;
     private Pattern getLangPattern;
     private Pattern matchSitePrefixPathPattern;
 
-    PathUrlLanguagePatternHandler(String sitePrefixPath) {
+    PathUrlLanguagePatternHandler(Lang defaultLang, ArrayList<Lang> supportedLangs, String sitePrefixPath) {
+        this.defaultLang = defaultLang;
+        this.supportedLangs = supportedLangs;
         this.sitePrefixPath = sitePrefixPath;
         this.getLangPattern = this.buildGetLangPattern(sitePrefixPath);
         this.matchSitePrefixPathPattern = this.buildMatchSitePrefixPathPattern(sitePrefixPath);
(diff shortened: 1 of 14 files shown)


src/test/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandlerTest.java
```diff
@@ -1,55 +1,97 @@
...
+        this.french = Lang.get("fr");
+
+        this.defaultLang = this.english;
+        this.supportedLangs = new ArrayList<Lang>();
+        this.supportedLangs.add(this.english);
+        this.supportedLangs.add(this.japanese);
+        this.supportedLangs.add(this.french);
+    }
+
+    private PathUrlLanguagePatternHandler createWithParams(String sitePrefixPath) {
```
bstewart00: wouldn't keep the WithParams in this name

src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
```diff
@@ -34,6 +40,15 @@ public boolean canInterceptUrl(String url) {
         return this.matchSitePrefixPathPattern.matcher(url).lookingAt();
     }
 
+    /*
+     * Redirect to same URL without language code if the language code
+     * found in the URL path is for default language
+     */
+    public boolean shouldRedirectToDefaultLang(String url) {
```
bstewart00: Don't see any tests for this 😭 
`shouldRedirectExplicitDefaultLangUrl` would be my choice of name. "Explicit default language" I felt describes this scenario a bit more.

Would potentially also remove this from the abstract base class and make it mandatory. In wovn.cs it is done for all patterns other than custom_domain, like subdomain should redirect en.domain.com to domain.com too

torkelbe: Thanks for the feedback! I'll try to adjust this accordingly.

Request handling seems to never have been tested very well, particularly for redirections and forwarding. The challenge is in mocking our environment. I'm still undecided in how to best add more robus testing, so for now I'm ok with moving forwards step by step, even if it's not perfect. That said, I'll look into adding something for this PR as well.

torkelbe: I changed the variable name and added unit tests for it. Thanks for the heads up on that. I'd like redirection to be tested better at a higher level, but that will be a separate effort.

For the default implementation of the function, I'd prefer to leave it like this for now since only path pattern has a custom implementation. We can discuss whether subdomain and query should also have default lang redirections, but I would prefer to not bundle such a change here.

PR #134, opened 2019-11-25
Change removeLang to convertToDefaultLang

torkelbe:
With Custom Domain, it does not always make sense to "remove lang code" to convert a URL to default language. Therefore, we change the name of `UrlLanguagePatternHandler.removeLang` to  `UrlLanguagePatternHandler.convertToDefaultLang`, and remove the language argument.

One consequence of this is that the function can only affect URL language codes that match a supported language. (This is why some language codes in tests need to change.)

Note that changes to the internals of the pattern handlers are deliberately kept to a minimum for this PR.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/SubdomainUrlLanguagePatternHandler.java
+++ b/src/main/java/com/github/wovnio/wovnjava/SubdomainUrlLanguagePatternHandler.java
@@ -19,7 +19,16 @@ class SubdomainUrlLanguagePatternHandler extends UrlLanguagePatternHandler {
         return (lang != null && this.supportedLangs.contains(lang)) ? lang : null;
     }
 
-    String removeLang(String url, String lang) {
+    String convertToDefaultLanguage(String url) {
+        Lang currentLang = this.getLang(url);
+        if (currentLang == null) {
+            return url;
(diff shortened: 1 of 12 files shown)


src/main/java/com/github/wovnio/wovnjava/SubdomainUrlLanguagePatternHandler.java
```diff
@@ -19,7 +19,16 @@ Lang getLang(String url) {
         return (lang != null && this.supportedLangs.contains(lang)) ? lang : null;
     }
 
-    String removeLang(String url, String lang) {
+    String convertToDefaultLanguage(String url) {
```
bstewart00: Could be a candidate for moving the implementation to the base class since it seems like it will always be the same, but will leave this up to you.

torkelbe: That could be possible, but it references `removeLang` which I've changed to a private method inside the implementations. I think I'll leave it as it is because I don't want to force all pattern handlers to have a `removeLang` method, and it way gives me better flexibility to improve the internals of path query and subdomain handlers.

PR #136, opened 2019-11-27
Support Custom Domain configuration

torkelbe:
## Custom Domain
Custom Domain is a URL pattern configuration alternative to path, query, and subdomain. It can specify the domain and path corresponding to each language independently.

### Example 
Example domain and path mapping to original language Japanese plus two target languages:
```
"www.site.co.jp"       -> Japanese
"www.site.com/english" -> English
"spain.site.com"       -> Spanish
```
For this example mapping, the URL `http://www.site.co.jp/about.html` will map to the three languages as follows
```
Japanese  -> "http://www.site.co.jp/about.html"
English   -> "http://www.site.com/english/about.html"
Spanish   -> "http://spain.site.com/about.html"
```
The corresponding `web.xml` configuration looks as follows
```xml
<filter>
  ...
  <init-param>
    <param-name>defaultLang</param-name>
    <param-value>ja</param-value>
  </init-param>
  <init-param>
    <param-name>supportedLangs</param-name>
    <param-value>ja,en,es</param-value>
  </init-param>
  <init-param>
    <param-name>urlPattern</param-name>
    <param-value>customDomain</param-value>
  </init-param>
  <init-param>
    <param-name>customDomainLangs</param-name>
    <param-value>www.site.co.jp:ja,www.site.com/english:en,spain.site.com:es</param-value>
  </init-param>
  ...
</filter>
```

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
+++ b/src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
@@ -12,7 +12,7 @@ class CustomDomainLanguage {
 
     public CustomDomainLanguage(String host, String path, Lang lang) {
         this.host = host;
-        this.path = path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
+        this.path = path;
         this.lang = lang;
     }
 
--- a/src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
+++ b/src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
@@ -2,8 +2,6 @@ package com.github.wovnio.wovnjava;
 
 import java.net.URL;
 import java.util.Arrays;
-import java.util.regex.Pattern;
-import java.util.regex.Matcher;
 
 class CustomDomainLanguage {
     public final String host;
--- a/src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
+++ b/src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
@@ -1,0 +1,41 @@
+package com.github.wovnio.wovnjava;
+
+import java.net.URL;
+import java.util.Arrays;
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
+
+class CustomDomainLanguage {
+    public final String host;
+    public final String path;
+    public final Lang lang;
+
+    public CustomDomainLanguage(String host, String path, Lang lang) {
+        this.host = host;
+        this.path = path.endsWith("/") ? path : path + "/";
+        this.lang = lang;
+    }
+
+    public boolean isMatch(URL url) {
+        if (url == null) return false;
+
+        return this.host.equalsIgnoreCase(url.getHost()) && this.pathIsEqualOrSubsetOf(this.path, url.getPath());
+    }
+
+    private boolean pathIsEqualOrSubsetOf(String basePath, String testPath) {
+        String[] basePathList = Arrays.stream(basePath.split("/")).filter(s -> !s.isEmpty()).toArray(String[]::new);
+        String[] testPathList = Arrays.stream(testPath.split("/")).filter(s -> !s.isEmpty()).toArray(String[]::new);
+
+        if (basePathList.length > testPathList.length) {
+            return false;
+        }
+
+        for (int i = 0; i < basePathList.length; i++) {
+            if (!basePathList[i].equalsIgnoreCase(testPathList[i])) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
--- a/src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
+++ b/src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
@@ -12,7 +12,7 @@ class CustomDomainLanguage {
 
     public CustomDomainLanguage(String host, String path, Lang lang) {
         this.host = host;
-        this.path = path.endsWith("/") ? path : path + "/";
+        this.path = path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
         this.lang = lang;
     }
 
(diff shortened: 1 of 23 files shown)


src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public boolean isMatch(URL url) {
+        if (url == null) return false;
+
+        return this.host.equalsIgnoreCase(url.getHost()) && this.pathIsEqualOrSubsetOf(this.path, url.getPath());
+    }
+
+    private boolean pathIsEqualOrSubsetOf(String basePath, String testPath) {
+        String[] basePathList = Arrays.stream(basePath.split("/")).filter(s -> !s.isEmpty()).toArray(String[]::new);
```
bstewart00: I noticed you used the Stream API here but not other places, for example CustomDomainLanguages etc that are basic for loops with filtering.
This is also somewhat related to the Java 6/backwards compatibility discussion though, as streams are a Java 8 feature.

src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguageValidator.java
```diff
@@ -0,0 +1,39 @@
...
+import java.net.URL;
+
+class CustomDomainLanguageValidator {
+    private CustomDomainLanguageValidator() {}
+
+    public static ValidationResult validate(ArrayList<CustomDomainLanguage> customDomainLanguageList, ArrayList<Lang> supportedLangs) {
+        ArrayList<Lang> includedLangs = new ArrayList<Lang>();
+        ArrayList<String> includedUrls = new ArrayList<String>();
+
+        for (CustomDomainLanguage customDomainLanguage : customDomainLanguageList) {
```
bstewart00: I think there's a `groupingBy` stream collector that could express this more concisely, but this is fine as well.

src/main/java/com/github/wovnio/wovnjava/CustomDomainUrlLanguagePatternHandler.java
```diff
@@ -0,0 +1,66 @@
...
+        URL url = getUrlObject(urlString);
+        if (url == null) return urlString;
+
+        CustomDomainLanguage currentCDL = this.customDomainLanguages.getCustomDomainLanguageByUrl(url);
+        if (currentCDL == null) return urlString;
+
+        CustomDomainLanguage targetCDL = this.customDomainLanguages.getCustomDomainLanguageByLang(lang);
+        if (targetCDL == null || currentCDL == targetCDL) return urlString;
+
+        Pattern p = Pattern.compile("^(" + currentCDL.path + ")");
```
bstewart00: No problems with trailing slashes etc with this regex? Haven't looked at the tests yet

src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguageSerializer.java
```diff
@@ -0,0 +1,55 @@
...
+    public static String serializeToJson(CustomDomainLanguages customDomainLanguages) {
+        if (customDomainLanguages == null) {
+            return "";
+        }
+
+        ArrayList<String> items = new ArrayList<String>();
+        for (CustomDomainLanguage cdl : customDomainLanguages.customDomainLanguageList) {
+            items.add(cdl.host + cdl.path + "/\":\"" + cdl.lang.code);
+        }
+        return "{\"" + String.join("\",\"", items) + "\"}";
```
bstewart00: I was kind of expecting there to be a standard json serializer in Java, but since this isn't too complicated it's _probably_ OK to manually do it here.

src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
torkelbe: Here I simply need to filter out empty elements of the array before doing further calculations, and this seemed to be the most concise way to do it.

I did not consider pre-java 8 support for these changes. But nice catch, we would have to change this filtering method for java 7 support. I'm torn on whether or not that means we should also change it for the java 8+ master branch.

src/main/java/com/github/wovnio/wovnjava/CustomDomainUrlLanguagePatternHandler.java
torkelbe: `CustomDomainLanguage.path` is always stored _without_ trailing slash. This is enforced by the Serializer when reading the configuration string. Maybe it should be enforced when initializing CustomDomainLanguage as well?

When we execute this regex, `CustomDomainLanguages.getCustomDomainLanguageByUrl` has already established that the paths match (followed by `/\/|$/`). So the regex just does a simple replace.

I believe the tests cover all the cases, but the various responsibilities are somewhat distributed among the `CustomDomainLanguage*` files.

src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguageSerializer.java
torkelbe: Yeah that's what I was thinking as well, and I looked for a good non-manual solution. But I believe it would require us to add another dependency to our `pom.xml`, which feels unnecessary for such a simple serialization. I think it's ok to keep it like this.

src/main/java/com/github/wovnio/wovnjava/CustomDomainUrlLanguagePatternHandler.java
bstewart00: I think I did the opposite in wovn.cs, where CustomDomainLanguage.Path was guaranteed to end with a trailing slash.

torkelbe: I ignored trailing slash to swap between `site.com` -> `site.com/ko` and back without additional logic for it, just as I would for `site.com/` -> `site.com/ko/`.

src/main/java/com/github/wovnio/wovnjava/CustomDomainLanguage.java
bstewart00: Definitely on board with using newer language features. Can address back-porting to older versions when it is needed.

PR #141, opened 2020-01-28
[Feature] Support langCodeAliases

torkelbe:
`langCodeAliases` lets you specify the language identifier for your supported languages.

> For example, the default language identifier for English is `en`, such that a URL for your page in English language may look like `http://site.com/en/page`.
> With `langCodeAliases`, you can change the language identifier to `us`, for example. The resulting URL would look like `http://site.com/us/page` instead.

See README for more details.

### Comments
This feature is relevant for path, query, and subdomain url patterns. As a result, the methods `getLang`, `convertToDefaultLanguage`, and `convertToTargetLanguage` has changed quite a bit for these pattern handlers.

`langCodeAliases` is a two-in-one feature from an implementation perspective. Allowing alias for a target language is a simple change, we only need to replace the default language code for identifying the language.

When an alias is set for default language, it changes the rules for interpreting requests. Request URLs without an explicit language identifier will now be ignored.

`getLang` has been changed to return `null` _only for input URLs that cannot be intercepted_. For any URL that WovnServletFilter should process, `getLang` will either return default language or a target language. (The last pattern handler, `CustomDomainUrlLanguagePatternHandler` already conforms to this behavior.)

With the new behavior of `getLang` , there is no longer need for the method `canInterceptUrl`. It has been removed.


bstewart00:
When I see frequent calls to `this.languageAliases.getLang`, it kind of suggests to me that it would be useful to have a "resolved language" concept that has the original language and the alias it mapped to. It would be a fairly foundational change though.

src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
```diff
@@ -6,64 +6,85 @@
...
 
-    private String removeLang(String url, String lang) {
-        if (lang.isEmpty()) return url;
+        String languageIdentifier = this.getLangMatch(url, this.getLangPattern);
+        Lang currentLang = this.languageAliases.getLang(languageIdentifier);
+        if (currentLang != null) {
+            String newUrl = this.removeLang(url, currentLang);
+            return this.insertLang(newUrl, targetLang);
+        } else if (this.languageAliases.hasAliasForDefaultLang) {
+            return url;
```
bstewart00: What case is this handling? I'm wondering if it is reachable in practice.

torkelbe:
```
defaultLang: "ja"
langCodeAlias: "ja": "japan"
--------------------
convertToTargetLanguage("http://site.com/global/page", <English>)
line 58) languageIdentifier   -> "global"
line 59) currentLang          -> null
```
Because language alias for default language is set, the input URL "http://site.com/global/page" falls outside of the accepted URLs, and we cannot modify it. So we return the input URL untouched at line 63.

-----
The helper function `getLangMatch` changed, and it's name may not be fitting anymore. I'll see if I can make this part less confusing.

torkelbe:
> When I see frequent calls to `this.languageAliases.getLang`, it kind of suggests to me that it would be useful to have a "resolved language" concept that has the original language and the alias it mapped to. It would be a fairly foundational change though.

I actually started towards a change like this but decided against it, at least for now. The reason is that I started having to change so many things unrelated to language aliases.

I agree that some files now fetch alias from language object and vice versa many times, and it could be elegant. But the benefit is that outside of the `UrlLanguagePatternHandler` for path, query and subdomain, the rest of the code can still handle the standard Lang objects as before, oblivious to aliases.

torkelbe:
I renamed the public methods of `LanguageAliases` to avoid naming confusion with pattern handlers etc
* `getLang` -> `getLanguageFromAlias`
* `getAlias` -> `getAliasFromLanguage`

src/main/java/com/github/wovnio/wovnjava/SubdomainUrlLanguagePatternHandler.java
```diff
@@ -1,55 +1,76 @@
...
-        this.supportedLangs = supportedLangs;
+        this.languageAliases = languageAliases;
         this.getLangPattern = this.buildGetLangPattern();
     }
 
     Lang getLang(String url) {
-        Lang lang = this.getLangMatch(url, this.getLangPattern);
-        return (lang != null && this.supportedLangs.contains(lang)) ? lang : null;
+        String languageIdentifier = this.resolvePatternMatch(url, this.getLangPattern);
+        Lang lang = this.languageAliases.getLanguageFromAlias(languageIdentifier);
```
bstewart00: Seems like this code doesn't handle if the language identifier isn't an alias? `getLanguageFromAlias` strictly only handles alias -> original

src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
bstewart00: I think I get it now. When the user sets a default lang alias, we assume they mean that the URLs to their source lang files MUST have the alias in it. So the language alias must always be resolvable by us.
`convertToTargetLanguage("http://site.com/global/page", <English>) => http://site.com/global/page`
without a default lang alias:
`convertToTargetLanguage("http://site.com/global/page", <English>) => http://site.com/en/global/page`

src/main/java/com/github/wovnio/wovnjava/SubdomainUrlLanguagePatternHandler.java
```diff
@@ -1,55 +1,76 @@
...
+        if (lang != null) {
+            return lang;
+        } else if (this.languageAliases.hasAliasForDefaultLang) {
+            return null;
```
bstewart00: It occurs to me that these special rules for default lang alias really only apply for path, but for consistency we have to do it I suppose.

src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
torkelbe: > When the user sets a default lang alias, we assume they mean that the URLs to their source lang files MUST have the alias in it.

Correct. 

> So if the default language alias isn't found in the URL, it must be malformed or something and we should do nothing.

It results in the same behavior as for site prefix path and custom domain when the input URL is something that does not match what we're allowed to change.

I'll try to add some comments, and maybe organise the test cases better.

src/main/java/com/github/wovnio/wovnjava/SubdomainUrlLanguagePatternHandler.java
torkelbe: `LanguageAliases` class will set up with the standard language code as the "alias" for all supported languages, unless another alias is configured using `langCodeAliases` setting.

It will currently allow only one string to identify a language, such that if "us" is configured as an alias for English, `"en"` will no longer be used to identify it.

bstewart00: ah. If all languages are considered their own alias by default then that makes sense.

src/main/java/com/github/wovnio/wovnjava/SubdomainUrlLanguagePatternHandler.java
torkelbe: I ignore `hasAliasForDefaultLang` for query url patterns because I did not feel it makes sense.

For subdomain, I feel like it is possible to let it work the same way as path (as is done here). With default language English using an alias `"us"`, it will mean that
* `http://us.site.com/*` is processed as request in default language
* `http://site.com/*` is identified as outside the scope of what wovnjava will handle, and is not processed
* `http://ja.site.com/*` is processed as request for content translated into Japanese

If we feel like it would be best to make this behavior different (like I did for query pattern), that's also ok.

bstewart00: I think that's OK. How about when there are no aliases at all, and they access "zz.site.com". Should it be going through the else block and returning default lang?

torkelbe: Yes, that sounds correct.

Without any aliases:
```
default language     -> "zz.site.com"
translated into thai -> "th.zz.site.com"
```

src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
torkelbe: I've added some comments to a few of the conditional branches that I think may be a bit confusing. I've also tried to name the tests more accurately.
I'm not sure it got a lot better, though.

src/main/java/com/github/wovnio/wovnjava/SubdomainUrlLanguagePatternHandler.java
```diff
@@ -1,55 +1,81 @@
...
-    String convertToTargetLanguage(String url, Lang lang) {
-        Lang currentLang = this.getLangMatch(url, this.getLangPattern);
-        if (currentLang != null && this.supportedLangs.contains(currentLang)) {
-            url = this.removeLang(url, currentLang.code);
+    String convertToTargetLanguage(String url, Lang targetLang) {
+        if (targetLang == this.defaultLang) {
+            return this.convertToDefaultLanguage(url);
+        }
+
+        String languageIdentifier = this.resolvePatternMatch(url, this.getLangPattern);
```
bstewart00: I think `findLanguageIdentifier` would be a better name

src/main/java/com/github/wovnio/wovnjava/PathUrlLanguagePatternHandler.java
bstewart00: I guess the most important thing is there are tests that fail if we tried to change this behavior. We want to make clear that we have decided that language aliases mean the un-aliased version is no longer recognized.

PR #143, opened 2020-03-06
Support for java6

iwai-wovn:
- Add compile scripts for jdk6 with maven
- Add CircleCI job for jdk6
- Add hello world sample of wovnjava jdk6
- Fix HtmlConverter.java cannot find symbol error


src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -162,8 +162,9 @@ private void appendHrefLang() {
     }
 
     private void replaceNodeToMarkerComment(Element element) {
-        Comment comment = new Comment(WOVN_MARKER_PREFIX + String.valueOf(markers.size()));
-        element.replaceWith(comment);
+        String commentData = WOVN_MARKER_PREFIX + String.valueOf(markers.size());
+        element.before("<!--" + commentData + "-->");
+        element.remove();
```
torkelbe: Nitpick: Because we no longer use the Comment class,  think we can remove its import also. (Line 9)

PR #146, opened 2020-03-29
[Fix] Changed character encoding only when converting HTML

iwai-wovn:
### Purpose/goal of PR

Don't need to set the character code the content-type is not HTML.
Fixed a bug that it is converted to image/png;charset=utf-8 when content-type is image/png.


bstewart00:
It would be ideal if you can add tests to WovnServletFilterTest.java that show that `image/png` does not get charset=utf-8 added to it. The problem is our `mockResponse` does not currently support making a fake response body which you might want to try.

iwai-wovn:
> It would be ideal if you can add tests to WovnServletFilterTest.java that show that `image/png` does not get charset=utf-8 added to it. The problem is our `mockResponse` does not currently support making a fake response body which you might want to try.
> 
> Even though there is currently a test case for images, it does not cover this scenario because the response body is empty.

That's right, I thought about adding testing, but I found it difficult to add testing, partly because the customer is rushing this release, so I'm going to go ahead and release it for the time being and leave the testing changes for later.

PR #147, opened 2020-03-31
Fixed the problem of converting everything to UTF-8

iwai-wovn:
If it is possible to convert to utf-8, all things such as images will be converted to utf-8, so confirm the Content-Type to convert only text or json.

Note: In the future, may need the option to exclude some URLs or Content-Types.

The test implementation to verify the HTTP header has not yet been implemented, so I will attach the local verification results for the time being.

It is noteworthy that no unnecessary character codes have been added to the Content-Type.

BEFORE: 
スクリーンショット 2020-03-31 22 25 13
AFTER: 
スクリーンショット 2020-03-31 22 25 23

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
@@ -25,11 +25,13 @@ class HtmlChecker {
     }
 
     public boolean isTextContentType(String type) {
+        String contentType = type.toLowerCase();
+
         return type == null
-            || type.toLowerCase().contains("text")
-            || type.toLowerCase().contains("html")
-            || type.toLowerCase().contains("application/json")
-            || type.toLowerCase().contains("application/javascript");
+            || contentType.contains("text")
+            || contentType.contains("html")
+            || contentType.contains("application/json")
+            || contentType.contains("application/javascript");
     }
 
     private boolean isAmp(String head) {
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
@@ -25,10 +25,12 @@ class HtmlChecker {
     }
 
     public boolean isTextContentType(String type) {
+        if (type == null)
+            return true;
+
         String contentType = type.toLowerCase();
 
-        return type == null
-            || contentType.contains("text")
+        return contentType.contains("text")
             || contentType.contains("html")
             || contentType.contains("application/json")
             || contentType.contains("application/javascript");
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
@@ -24,6 +24,14 @@ class HtmlChecker {
         return isHtml(head) && !isAmp(head);
     }
 
+    public boolean isTextContentType(String type) {
+        return type == null
+            || type.toLowerCase().contains("text")
+            || type.toLowerCase().contains("html")
+            || type.toLowerCase().contains("application/json")
+            || type.toLowerCase().contains("application/javascript");
+    }
+
     private boolean isAmp(String head) {
         Element element = Jsoup.parse(head).getElementsByTag("html").first();
         return element != null && (element.hasAttr("amp") || element.hasAttr("⚡"));
--- a/src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
@@ -89,8 +89,9 @@ public class WovnServletFilter implements Filter {
             wovnRequest.getRequestDispatcher(headers.getCurrentContextUrlInDefaultLanguage().getPath()).forward(wovnRequest, wovnResponse);
         }
 
-        String originalBody = wovnResponse.toString();
-        if (originalBody != null) {
+        if (htmlChecker.isTextContentType(response.getContentType())) {
+            String originalBody = wovnResponse.toString();
+
             // text
             String body = null;
             if (htmlChecker.canTranslate(response.getContentType(), originalBody)) {
@@ -98,13 +99,12 @@ public class WovnServletFilter implements Filter {
                 Api api = new Api(settings, headers, requestOptions, responseHeaders);
                 Interceptor interceptor = new Interceptor(headers, settings, api, responseHeaders);
                 body = interceptor.translate(originalBody);
-
-                wovnResponse.setCharacterEncoding("utf-8");
             } else {
                 // css, javascript or others
                 body = originalBody;
             }
             wovnResponse.setContentLength(body.getBytes().length);
+            wovnResponse.setCharacterEncoding("utf-8");
             PrintWriter out = response.getWriter();
             out.write(body);
             out.close();
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
-    <version>1.1.1-java7</version>
+    <version>1.1.2-java7</version>
     <url>https://github.com/WOVNio/wovnjava</url>
 
     <licenses>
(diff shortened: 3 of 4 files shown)


src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
```diff
@@ -89,22 +89,22 @@ private void tryTranslate(Headers headers, RequestOptions requestOptions, HttpSe
             wovnRequest.getRequestDispatcher(headers.getCurrentContextUrlInDefaultLanguage().getPath()).forward(wovnRequest, wovnResponse);
         }
 
-        String originalBody = wovnResponse.toString();
-        if (originalBody != null) {
+        if (htmlChecker.isTextContentType(response.getContentType())) {
```
bstewart00, line 92: Why is `htmlChecker.canTranslateContentType` not good enough? We only care about translating html anyway

bstewart00: Or is there problems if text content goes through the binary path?

iwai-wovn: A big reason why I implemented it separately is because I thought javascript should also be converted to UTF-8, because in my experience, there have been cases where javascript and HTML character codes don't match, which leads to a glitch.

Therefore, I thought that what can be translated and what should be converted to utf-8 are two different things, so I implemented the method separately.

Incidentally, wovnResponse.toString() is the return value of the implicitly converted utf-8 string.

bstewart00: oh, so everything is already converted to UTF-8 anyway in `wovnResponse.toString()`. I think wovnjava isn't totally correct still. Maybe someone uses SHIFT-JIS for their CSS files which would be broken. This is an improvement though.

src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
```diff
@@ -24,6 +24,14 @@ public boolean canTranslateContent(String html) {
         return isHtml(head) && !isAmp(head);
     }
 
+    public boolean isTextContentType(String type) {
+        return type == null
+            || type.toLowerCase().contains("text")
```
bstewart00, line 29: put `type.toLowerCase()` in a variable to avoid repeating it, even if it makes the code less nice.

src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
iwai-wovn: Should CSS not be converted to utf-8?

bstewart00: it doesn't necessarily have to be UTF-8. I don't think we support it yet though

iwai-wovn: Ah, maybe it doesn't support rewriting @charset in CSS, so converting it to utf-8 would be a problem. You mean?

bstewart00: Yes, you can use `@charset` to make it something other than utf-8. None of our backends would handle this currently though.

iwai-wovn: I see, CSS should be left out of the UTF-8 conversion.

bstewart00: I'm saying you should not worry about this now. If we use `wovnResponse.toString()` then we have to set utf-8 for all text responses. Ideally, we would use the same encoding that came from the original response instead of assuming utf-8

iwai-wovn: Maybe I should change the name of the method, too, like `canUtf8ConvertableContentType()`?

src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
```diff
@@ -24,6 +24,16 @@ public boolean canTranslateContent(String html) {
         return isHtml(head) && !isAmp(head);
     }
 
+    public boolean isTextContentType(String type) {
+        String contentType = type.toLowerCase();
+
```
iwai-wovn, line 29: @bstewart00  I think of it this way, but what do you think?

```suggestion
    public boolean canUtf8ConvertableContentType(String type) {
        String contentType = type.toLowerCase();

        if (contentType.contains("text/css")) {
            return false;
        }

```

src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
bstewart00: I think the current name is OK. `wovnResponse.toString()` doesn't actually convert non-utf8 to utf8 either

src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
bstewart00: This is too difficult to handle now, don't worry about it 😄 
we _do_ need to use utf-8 for css if the css was indeed utf-8. But if they used SHIFT-JIS for example, we would need to use that. However, wovnjava assumes all text responses are UTF-8, so we can't really do this now.

iwai-wovn: Well, thank you!
I'll leave you with just a comment as a future concern.

src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
iwai-wovn: Isn't this line converting to Utf-8?
https://github.com/WOVNio/wovnjava/blob/master/src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java#L49

bstewart00: If you look at how it is implemented, it is just doing `new String(data, "utf-8)` which is only interpreting the bytes as utf-8, not actually converting.

iwai-wovn: Oh, I hadn't seen the implementation. If you do so, the character encoding specified by `OutputStreamWriter` will be forced to utf8, so it's better to get and set the character encoding of the original string, right?
https://github.com/WOVNio/wovnjava/blob/master/src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java#L70

iwai-wovn: No, I guess it's better to keep it that way.

bstewart00: Something like that, ideally

Soichiro725:
Can we trust `Content-Type` ?
WOVN.php is doing same things by checking requested URL.
https://github.com/WOVNio/WOVN.php/blob/ea1b5a99b34a70b01779218b03c97b128b16364f/src/wovnio/wovnphp/Utils.php#L6

bstewart00:
> Can we trust `Content-Type` ?
> WOVN.php is doing same things by checking requested URL.
> https://github.com/WOVNio/WOVN.php/blob/ea1b5a99b34a70b01779218b03c97b128b16364f/src/wovnio/wovnphp/Utils.php#L6

What wovn.php does is more dangerous I think. Any URL can be any content type, there doesn't have to be a file extension

iwai-wovn:
@Soichiro725 
> Can we trust `Content-Type` ?
> WOVN.php is doing same things by checking requested URL.

WovnJava also uses URLs to make decisions, but the problem this time was when the image was dynamically output from the application.
In that case, the URL has no extension, so it can only be judged from the Content-Type.

src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
```diff
@@ -89,22 +89,22 @@ private void tryTranslate(Headers headers, RequestOptions requestOptions, HttpSe
             wovnRequest.getRequestDispatcher(headers.getCurrentContextUrlInDefaultLanguage().getPath()).forward(wovnRequest, wovnResponse);
         }
 
-        String originalBody = wovnResponse.toString();
-        if (originalBody != null) {
```
Soichiro725, line 93: IS that OK to remove this condition?

iwai-wovn: If you check the implementation of this code, there are essentially no cases where the constructor can return a null, as in Java, it is not allowed to return a null.
Therefore, the original code is not working at all.
https://github.com/WOVNio/wovnjava/blob/master/src/main/java/com/github/wovnio/wovnjava/Utf8.java#L34

iwai-wovn:
thanks reviewers!!

PR #150, opened 2020-04-29
Added test for Region Code

iwai-wovn:
* The URL pattern is now always case-insensitive.
* Supported language keys are in lower case only.
* Add test for Chinese lang code


src/main/java/com/github/wovnio/wovnjava/LanguageAliases.java
```diff
@@ -26,10 +26,17 @@ public String getAliasFromLanguage(Lang lang) {
     }
 
     public Lang getLanguageFromAlias(String alias) {
+        return getLanguageFromAlias(alias, false);
```
bstewart00, line 29: There's no need for two functions. It should _always_ be case insensitive

src/main/java/com/github/wovnio/wovnjava/Lang.java
```diff
@@ -57,22 +57,22 @@
...
+        map.put("es-co",  new Lang("es-CO", "es-CO", "Spanish (Colombia)", "Español (Colombia)"));
+        map.put("es-cr",  new Lang("es-CR", "es-CR", "Spanish (Costa Rica)", "Español (Costa Rica)"));
+        map.put("es-hn",  new Lang("es-HN", "es-HN", "Spanish (Honduras)", "Español (Honduras)"));
         map.put("es-419", new Lang("es-419", "es-419", "Spanish (Latin America)", "Español (Latinoamérica)"));
-        map.put("es-MX",  new Lang("es-MX", "es-MX", "Spanish (Mexico)", "Español (México)"));
-        map.put("es-PE",  new Lang("es-PE", "es-PE", "Spanish (Peru)", "Español (Perú)"));
-        map.put("es-ES",  new Lang("es-ES", "es-ES", "Spanish (Spain)", "Español (España)"));
-        map.put("es-US",  new Lang("es-US", "es-US", "Spanish (United States)", "Español (Estados Unidos)"));
-        map.put("es-UY",  new Lang("es-UY", "es-UY", "Spanish (Uruguay)", "Español (Uruguay)"));
-        map.put("es-VE",  new Lang("es-VE", "es-VE", "Spanish (Venezuela)", "Español (Venezuela)"));
```
bstewart00, line 75: This map requiring lowercase keys is a bit fragile. If you want, you could refactor it to be more like this:
```
        ArrayList<Lang> langs = new ArrayList<Lang>();
        langs.add(new Lang("ar", "ar", "Arabic", "ﺎﻠﻋﺮﺒﻳﺓ"));
        //...
        HashMap<String, Lang> map = new HashMap<String, Lang>();
        for(Lang lang: langs){
            map.put(lang.code.toLowerCase(), lang);
         } 
        LANG = Collections.unmodifiableMap(map);
```

src/test/java/com/github/wovnio/wovnjava/QueryUrlLanguagePatternHandlerTest.java
```diff
@@ -59,8 +85,20 @@ public void testGetLang__MatchingQuery__ValidSupportedLang__ReturnTargetLangObje
         assertEquals(this.french, sut.getLang("en.site.com/es/page/index.html?wovn=fr"));
     }
 
+    public void testGetLang__MatchingQuery__ValidSupportedLang__ReturnTargetLangObjectOfChinese() {
```
bstewart00, line 88: These tests are hard to name, but I would call this one `testGetLang__MatchingQuery__ValidSupportedRegionalLang__ReturnTargetLangObject`
just so it isn't specifically about Chinese.

src/test/java/com/github/wovnio/wovnjava/QueryUrlLanguagePatternHandlerTest.java
```diff
@@ -47,8 +61,20 @@ public void testGetLang__NonMatchingQuery__ReturnDefaultLang() {
         assertEquals(this.defaultLang, sut.getLang("http://site.com?wovn="));
     }
 
+    public void testGetLang__NonMatchingQuery__ReturnDefaultLangByChinese() {
```
bstewart00, line 64: I would say the "NonMatching" tests don't need separate test cases for zh-CHT. These tests just assert that the default language is returned so there isn't any special behavior this is testing.

src/test/java/com/github/wovnio/wovnjava/QueryUrlLanguagePatternHandlerTest.java
bstewart00: Actually, you could just put these cases inside of the test above it as well. Similar to 
`testGetLang__MatchingSubdomain__ValidSupportedLang__ReturnTargetLangObject`

src/main/java/com/github/wovnio/wovnjava/LanguageAliases.java
iwai-wovn: It should be case-insensitive in Subdomain, but shouldn't it be case-insensitive in path and query patterns?

bstewart00: We want path, subdomain, query patterns to all be case-insensitive. So always use `equalsIgnoreCase`

src/main/java/com/github/wovnio/wovnjava/Lang.java
```diff
@@ -1,88 +1,94 @@
...
+        langs.add(new Lang("sw", "sw", "Swahili", "Kiswahili"));
+        langs.add(new Lang("sv", "sv", "Swedish", "Svensk"));
+        langs.add(new Lang("tl", "tl", "Tagalog", "Tagalog"));
+        langs.add(new Lang("th", "th", "Thai", "ภาษาไทย"));
+        langs.add(new Lang("hi", "hi", "Hindi", "हिन्दी"));
+        langs.add(new Lang("tr", "tr", "Turkish", "Türkçe"));
+        langs.add(new Lang("uk", "uk", "Ukrainian", "Українська"));
+        langs.add(new Lang("ur", "ur", "Urdu", "اردو"));
+        langs.add(new Lang("vi", "vi", "Vietnamese", "Tiếng Việt"));
+        langs.add(new Lang("tl", "tl", "Tagalog", "Tagalog"));
```
bstewart00, line 86: There is a duplicate the Tagalog.

PR #153, opened 2020-05-12
Add `ignorePaths` option feature

iwai-wovn:
### Purpose/goal of PR (Issue #)
Add the ability for users to ignore some paths.
Users can set ignore paths in web.xml file as follows:
```
<init-param>
  <param-name>ignorePaths</param-name>
  <param-value>/admin,/wp-admin</param-value>
</init-param>
```
### Comments
Ignore paths is a forward match. And `sitePrefixPath` is not taken into account. Therefore, if you set a `sitePrefixPath`, you will need to add a prefix path to the `ignorePaths` as well.

### DIFF
--- a/README.en.md
+++ b/README.en.md
@@ -221,6 +221,9 @@ With this configuration, WOVN.java will ignore the following URLs
 https://my-wesite.com/admin
 https://my-wesite.com/admin/
 https://my-website.com/admin/plugin.html
+https://my-wesite.com/wp-admin
+https://my-wesite.com/wp-admin/
+https://my-website.com/wp-admin/anypages
 ```
 but allow the following
 ```Text
--- a/README.en.md
+++ b/README.en.md
@@ -202,7 +202,34 @@ Configure WovnServletFilter to determine request path and query from the same HT
 _The sample request header shown above was referenced from the following site:_
 https://coderwall.com/p/jhkw7w/passing-request-uri-into-request-header
 
-### 2.7. ignoreClasses
+### 2.7. ignorePaths
+
+A comma-separated list of URL path for which you would like WOVN to not translation content withing given directories.
+
+Ignored paths and their contents will not be processed by WovnServletFilter, and will not be sent to Wovn.io for translation. 
+
+For instance, if you want to not translation the admin directory of your website, you should configure as below.
+
+```XML
+<init-param>
+  <param-name>ignorePaths</param-name>
+  <param-value>/admin,/wp-admin</param-value>
+</init-param>
+```
+With this configuration, WOVN.java will ignore the following URLs
+```Text
+https://my-wesite.com/admin
+https://my-wesite.com/admin/
+https://my-website.com/admin/plugin.html
+```
+but allow the following
+```Text
+https://my-website.com/index.html
+https://my-website.com/user/admin
+https://my-website.com/adminpage
+```
+
+### 2.8. ignoreClasses
 
 A comma-separated list of HTML classes for which you would like WOVN to skip the elements of.
 (This setting is used to prevent confidential data contained in the page to be translated from being sent to WOVN)
@@ -229,7 +256,7 @@ Including three classes, `email-address-element`, `my-secret-class`, and `noshow
 </init-param>
 ```
 
-### 2.8. enableFlushBuffer
+### 2.9. enableFlushBuffer
 A flag to adjust the behavior of `ServletResponse.flushBuffer()`.
 
 This parameter is set to `false` by default (recommended).
@@ -238,7 +265,7 @@ When `enableFlushBuffer` is set to `false`, WovnServletFilter will capture calls
 immediately writing content to the client. Only when the complete HTML response is ready will the filter translate the content
 and send it to the client. This is necessary in order to translate the content properly.
 
-### 2.9. sitePrefixPath
+### 2.10. sitePrefixPath
 
 This parameter lets you set a prefix path to use as an anchor for which WOVN will translate pages. With this setting, WOVN will only translate pages that match the prefix path, and the path language code will be added _after_ the prefix path.
 
@@ -268,7 +295,7 @@ Furthermore, it is highly recommended to also configure your `web.xml` with a co
 </filter-mapping>
 ```
 
-### 2.10. langCodeAliases
+### 2.11. langCodeAliases
 
 This setting lets you specify the language identifier for your supported languages.
 
@@ -303,7 +330,7 @@ To illustrate, here is an example:
 
 Achieve this result by configuring `jp` as an alias for Japanese.
 
-### 2.11. customDomainLangs
+### 2.12. customDomainLangs
 
 This setting lets you define the domain and path that corresponds to each of your supported languages.
 
@@ -344,7 +371,7 @@ If this setting is used, each language declared in `supportedLangs` must be give
 Lastly, the path declared for your original language must match the structure of the underlying web server.
 In other words, you cannot use this setting to change the request path of your content in original language.
 
-### 2.12. debugMode
+### 2.13. debugMode
 
 A flag to enable extra debugging features.
 
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
-    <version>1.1.3</version>
+    <version>1.2.3</version>
     <url>https://github.com/WOVNio/wovnjava</url>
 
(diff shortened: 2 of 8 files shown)


src/test/java/com/github/wovnio/wovnjava/HeadersTest.java
```diff
@@ -240,24 +240,60 @@ public void testLocationWithSitePrefixPath() throws ConfigurationError {
...
+
+        h = makeHeaderWithSitePrefixPathAndIgnorePaths("/city/wp-admin", "city", "/admin,/wp-admin");
+        assertEquals(true, h.getIsValidRequest());
+
+        h = makeHeaderWithSitePrefixPathAndIgnorePaths("/city/wp-admin", "city", "/city/admin,/city/wp-admin");
         assertEquals(false, h.getIsValidRequest());
     }
 
-    private Headers makeHeaderWithSitePrefixPath(String requestPath, String sitePrefixPath) throws ConfigurationError {
+    private Headers makeHeaderWithSitePrefixPathAndIgnorePaths(String requestPath, String sitePrefixPath, String ignorePaths) throws ConfigurationError {
```
bstewart00, line 291: I would be OK renaming this to just `createHeaders`

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -52,7 +54,7 @@
 
         this.shouldRedirectExplicitDefaultLangUrl = this.urlLanguagePatternHandler.shouldRedirectExplicitDefaultLangUrl(clientRequestUrl);
 
-        this.isValidRequest = this.requestLang != null && this.urlContext != null;
+        this.isValidRequest = this.requestLang != null && this.urlContext != null && currentContextUrlIgnoredPath == false;
```
bstewart00, line 57: `UrlContext` shouldn't really care about ignore paths. It would be better as a method in headers:
`this.isIgnoredPath(this.urlContext.getPath())`

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -127,4 +128,29 @@ public boolean getIsValidRequest() {
         hreflangs.put(hreflangCodeDefaultLang, urlDefaultLang);
         return hreflangs;
     }
+
+    private boolean isIgnoredPath(String contextPath) {
+        ArrayList<String> ignorePaths = this.settings.ignorePaths;
+
+        if (ignorePaths.isEmpty()) {
```
bstewart00, line 135: this check isn't needed

src/test/java/com/github/wovnio/wovnjava/HeadersTest.java
```diff
@@ -240,24 +240,60 @@ public void testLocationWithSitePrefixPath() throws ConfigurationError {
...
-        h = makeHeaderWithSitePrefixPath("/global/ja/foo", "global");
+        h = createHeaders("/global/ja/foo", "global", "");
         assertEquals(true, h.getIsValidRequest());
 
-        h = makeHeaderWithSitePrefixPath("/ja/global/foo", "global");
+        h = createHeaders("/ja/global/foo", "global", "");
+        assertEquals(false, h.getIsValidRequest());
+
+
+        h = createHeaders("/", "", "/admin,/wp-admin");
```
bstewart00, line 256: move the ignore class test cases into a separate test: `testGetIsValidRequest__withIgnoredPaths`

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -127,4 +128,29 @@ public boolean getIsValidRequest() {
...
+            return false;
+        }
+
+        for (String ignorePath : ignorePaths) {
+            if (ignorePath.isEmpty()) {
```
bstewart00, line 140: I would not bother with this check. Or in Settings.java you could make a `normalizeIgnorePaths` method where you can remove the invalid paths, and maybe insert missing slashes

pom.xml
```diff
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
-    <version>1.1.3</version>
+    <version>1.2.3</version>
```
bstewart00, line 10: version number should be reset to 1.2.0

src/test/java/com/github/wovnio/wovnjava/HeadersTest.java
```diff
@@ -240,24 +240,63 @@ public void testLocationWithSitePrefixPath() throws ConfigurationError {
...
+        h = createHeaders("/user/admin", "", "/admin,/wp-admin");
+        assertEquals(true, h.getIsValidRequest());
+
+        h = createHeaders("/adminpage", "", "/admin,/wp-admin");
+        assertEquals(true, h.getIsValidRequest());
+
+        h = createHeaders("/admin", "", "/admin,/wp-admin");
+        assertEquals(false, h.getIsValidRequest());
+
+        h = createHeaders("/wp-admin/", "", "/admin,/wp-admin");
```
bstewart00, line 271: Add another case for /wp-admin/page for when there are more paths at the end.

pom.xml
iwai-wovn: Oh, sorry...

src/test/java/com/github/wovnio/wovnjava/SettingsTest.java
```diff
@@ -328,6 +330,31 @@ public void testSnippetUrl__DevelopmentMode__DefaultValue() throws Configuration
...
+        }});
+        Settings s = new Settings(config);
+        ArrayList<String> emptyArrayList = new ArrayList<String>();
+        assertEquals(emptyArrayList, s.ignorePaths);
+    }
+
+    public void testIgnorePaths__NormalizedValues() throws ConfigurationError {
+        Settings s;
+
+        s = createSettings(new HashMap<String, String>() {{
```
bstewart00, line 345: make these separate test methods.
`testIgnorePaths__emptyValues__areExcluded`
`testIgnorePaths__pathsWithoutLeadingSlash__slashAdded`
`testIgnorePaths__pathsWithTrailingSlash__slashRemoved`

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -127,4 +128,22 @@ public boolean getIsValidRequest() {
...
+
+    private boolean isIgnoredPath(String contextPath) {
+        ArrayList<String> ignorePaths = this.settings.ignorePaths;
+
+        for (String ignorePath : ignorePaths) {
+
+            String prefixPath = ignorePath;
+            String prefixPathTrailingSlash = ignorePath + "/";
+
+            if (contextPath.toLowerCase().equals(prefixPath)) {
```
bstewart00, line 140: Do `contextPath = contextPath.toLowerCase()` at the top.
and then you can do
```
return contextPath.equals(prefixPath) || contextPath.startsWith(prefixPathTrailingSlash)
```

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -127,4 +128,20 @@ public boolean getIsValidRequest() {
...
+        ArrayList<String> ignorePaths = this.settings.ignorePaths;
+
+        contextPath = contextPath.toLowerCase();
+
+        for (String ignorePath : ignorePaths) {
+
+            String prefixPath = ignorePath;
+            String prefixPathTrailingSlash = ignorePath + "/";
+
+            if (contextPath.equals(prefixPath) || contextPath.startsWith(prefixPathTrailingSlash))
```
bstewart00, line 142: It's a bit redundant to do `if(expression) return true`. Just do `return expression`

src/main/java/com/github/wovnio/wovnjava/Headers.java
bstewart00: I forgot that this is inside of a loop, so what you did is correct

README.en.md
```diff
@@ -202,7 +202,34 @@ Configure WovnServletFilter to determine request path and query from the same HT
...
+<init-param>
+  <param-name>ignorePaths</param-name>
+  <param-value>/admin,/wp-admin</param-value>
+</init-param>
+```
+With this configuration, WOVN.java will ignore the following URLs
+```Text
+https://my-wesite.com/admin
+https://my-wesite.com/admin/
+https://my-website.com/admin/plugin.html
```
Soichiro725, line 223: The example also has `/wp-admin`.
So, I can understand easily, if there are examples with `/wp-admin`.

PR #159, opened 2020-10-07
Add TLS 1.2 support for java 6

bstewart00:
The recommended method is to use the BouncyCastle library that provides an independent implementation of Java's security and SSL library. https://www.bouncycastle.org/java.html

The basic step is doing `con.setSSLSocketFactory(new SSLSocketFactoryImpl());` to replace the default with BouncyCastle's socket factory. `Security.addProvider(new BouncyCastleProvider());` is needed to provide a working SecureRandom implementation used by the socket. wovnjava only makes a single request to html swapper, so it is only needed in one place.

I tested this by temporarily changing the code to make requests to https://tls-v1-2.badssl.com:1012/ which is a test site that _only_ uses TLS 1.2. Before this change, you would get a SSL handshake error.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Api.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Api.java
@@ -17,12 +17,22 @@ import java.util.zip.GZIPOutputStream;
 import java.util.LinkedHashMap;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.security.Security;
 
 import javax.xml.bind.DatatypeConverter;
+import javax.net.ssl.*;
 
 import net.arnx.jsonic.JSON;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.jsse.provider.SSLSocketFactoryImpl;
 
 class Api {
+	static {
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -1,26 +1,37 @@
...
 import javax.xml.bind.DatatypeConverter;
 
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.jsse.provider.SSLSocketFactoryImpl;
+
 import net.arnx.jsonic.JSON;
 
 class Api {
+	static {
+		if (Security.getProvider(BouncyCastleProvider.PROVIDER_NAME) == null) {
```
suplo, line 30: tab vs space :see_no_evil:

bstewart00: ack, I think it's my first time editing java code in VSCode so my settings may have been off.

aw-leigh (approved):

iwai-wovn (approved):

iwai-wovn:
@bstewart00 
Please create your release when you have time.
https://github.com/WOVNio/equalizer/wiki/Publish-wovnjava

PR #164, opened 2020-11-05
Fix/correct handling of wovn ignore

zeyuwu:
After this fix, all comments inserted by wovnjava into the HTML preserves the original tag information.

HTML:                    `<div><p class="wovn-ignore">message</p></div>`
Original behavior:  `<div><!--wovn-marker-0--></div>`
New behavior:       `<div><p class="wovn-ignore"><!--wovn-marker-0></p></div>`

for `<input>` tags, they are converted to `<input wovn-marker-2="">`, discarding everything else inside the `<input>` tag.

The implementation follows closely how WOVN.php does it. Overall idea being record comment text as keys for later lookup during restoring process. The restoring process is a simple string replace.


src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -44,20 +47,29 @@ String convert(String lang) {
...
-        sb.append(list[0]);
-        for (int i=1; i<list.length; i++) {
-            String fragment = list[i];
-            String commentSuffix = "-->";
-            int suffixOffset = fragment.indexOf(commentSuffix);
-            String indexString = fragment.substring(0, suffixOffset);
-            int index = Integer.parseInt(indexString);
-            sb.append(markers.get(index));
-            sb.append(fragment.substring(suffixOffset + commentSuffix.length()));
+        Document document = Jsoup.parse(html);
```
bstewart00, line 50: The point of the html string is we don't want to re-parse the document again, ideally.

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -168,9 +180,17 @@ private void appendHrefLang() {
     }
 
     private void replaceNodeToMarkerComment(Element element) {
-        Comment comment = new Comment(WOVN_MARKER_PREFIX + String.valueOf(markers.size()));
-        element.replaceWith(comment);
-        markers.add(restore(element.outerHtml())); // restore original text if element has marker
+        int commentId = this.markerManager.getCommentId();
+        Element newElement = restore(element);
+        if (this.markerManager.addComment(newElement, String.valueOf(commentId)) != commentId) {
+            throw new ConcurrentModificationException("Invlid Comment ID. CommentManger is not thread-safe.");
```
bstewart00, line 186: I would not bother with this check. We are not creating any threads, so you may assume a request is processed consistently within a single thread.

src/main/java/com/github/wovnio/wovnjava/HtmlMarkerManager.java
```diff
@@ -0,0 +1,24 @@
+package com.github.wovnio.wovnjava;
+
+import java.util.HashMap;
+
+import org.jsoup.nodes.Element;
+
+class HtmlMarkerManager {
```
bstewart00, line 7: I would recommend just straight-up copying the HtmlReplaceMarker implementation from the other backends.

bstewart00: wovnjava notably does not support the "attribute markers" (addValue in other backends), which would not be supported by this implementation as-is.

bstewart00: Actually, it does _attempt_ to do this in `replaceForm`, but it just replaces the whole `<input type="hidden">` node itself with a comment. Other backends just replace the value attribute.

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
zeyuwu: Ok.

src/main/java/com/github/wovnio/wovnjava/HtmlMarkerManager.java
zeyuwu: Got it.

zeyuwu:
Changed the implementation, now it should be very similar to how other libraries work.

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -119,7 +107,13 @@ private void removeForm() {
         for (Element element : elements) {
             String type = element.attr("type");
             if (type != null && type.toLowerCase().equals("hidden")) {
-                replaceNodeToMarkerComment(element);
+                String original = element.outerHtml();
```
bstewart00, line 110: This doesn't look correct 😉 Why is it removing attributes? The expected behavior is
```
<input type="hidden" value="some dynamic value">
->
<input type="hidden" value="__wovn__mark-3">
```

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -168,9 +162,9 @@ private void appendHrefLang() {
     }
 
     private void replaceNodeToMarkerComment(Element element) {
-        Comment comment = new Comment(WOVN_MARKER_PREFIX + String.valueOf(markers.size()));
-        element.replaceWith(comment);
-        markers.add(restore(element.outerHtml())); // restore original text if element has marker
+        String commentKey = htmlReplaceMarker.addCommentValue(htmlReplaceMarker.revert(element.html()));
```
Soichiro725, line 165: Is `htmlReplaceMarker.revert` needed?

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -168,9 +162,9 @@ private void appendHrefLang() {
...
+        element.html("");
```
bstewart00, line 166: are you sure you can't simply do `element.html('<!-- foo -->')`?

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
zeyuwu: Just values? I thought the idea is that it removes as much "dynamic" content as possible for better caching performance. Other than `values`, there are also stuff like `csrf_token` that needs to be removed. So I implemented this as removing all possible attributes.

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
zeyuwu: Yes, otherwise depending on the DOM traversal order, an `Element` might already have its child element commented out. We don't want that to happen.

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
zeyuwu: Nope, tried that. Raw `<!-- foo -->` isn't liked by jsoup, the `<` and `>` get's HTML encoded.

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
Soichiro725: I got it.
So, wovnjava is doing useless marking and useless reverting 😢 
We can leave it now.

zeyuwu: Yeah. Sometimes you see only 3 comments on the final HTML but they are number 7, number 9, number 11 etc. It is because of that.

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
bstewart00: I would prefer if you just follow the implementation of the other backends for consistency

zeyuwu: Got it. Will change it to values only.

src/main/java/com/github/wovnio/wovnjava/HtmlReplaceMarker.java
```diff
@@ -0,0 +1,39 @@
+package com.github.wovnio.wovnjava;
+
+import java.util.HashMap;
+
+class HtmlReplaceMarker {
+    private final String WOVN_MARKER_PREFIX = "wovn-marker-";
+    private final HashMap<String, String> mappedValues;
```
Soichiro725, line 7: I'm not familiar with Java.
Is `final` correct in this case?

zeyuwu: `final` mean the reference to `mappedValues` cannot change. But the contents can change, as long as it's the same HashMap object. So it is ok.

zeyuwu: The `final` for the `String` means `WOVN_MARKER_PREFIX` is a constant, but that is because Strings are immutable in Java. When you try to change a String you must create a new one and therefore the reference has to change.

bstewart00: This is a best practice in java, similar to `readonly` in C# for example.

Soichiro725 (approved):

zeyuwu:
There are two other PRs for Java 6 and Java 7, please don't forget to approve those as well. 🙇‍♂️

bstewart00 (approved):
Looks good.

PR #165, opened 2020-11-05
Fix/correct handling of wovn ignore java6

zeyuwu:
Same as #164,  fixes incorrect handling of wonv-ignore comments.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
@@ -105,14 +105,13 @@ class HtmlConverter {
         for (Element element : elements) {
             String type = element.attr("type");
             if (type != null && type.toLowerCase().equals("hidden")) {
-                String original = element.outerHtml();
-                Attributes attributes = element.attributes().clone();
-                for (Attribute attr: attributes) {
-                    element.removeAttr(attr.getKey());
+                if (element.hasAttr("value")) {
+                    String original = element.attr("value");
+                    String key = htmlReplaceMarker.generateKey();
+                    element.removeAttr("value").removeAttr("VALUE");
+                    element.attr("value", key);
+                    htmlReplaceMarker.addValue(key, original);
                 }
-                element.removeAttr("name");  // "name" doesn't show up in Element.attributes()
-                element.attr(htmlReplaceMarker.generateKey(), "");
-                htmlReplaceMarker.addValue(element.outerHtml(), original);
             }
         }
     }
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
@@ -3,22 +3,24 @@ package com.github.wovnio.wovnjava;
 import java.lang.StringBuilder;
 import java.util.Map;
 import java.util.HashMap;
-import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -116,7 +105,13 @@ private void removeForm() {
         for (Element element : elements) {
             String type = element.attr("type");
             if (type != null && type.toLowerCase().equals("hidden")) {
-                replaceNodeToMarkerComment(element);
+                if (element.hasAttr("value")) {
+                    String original = element.attr("value");
+                    String key = htmlReplaceMarker.generateKey();
+                    element.removeAttr("value").removeAttr("VALUE");
```
Soichiro725, line 111: Does only Java6 need `element.removeAttr("value").removeAttr("VALUE");` ?

zeyuwu: Yeah, it seems that the `jsoup` library we use for Java6 is case-sensitive when it comes to `setAttr()`, so this is done to make sure we set the correct one.

bstewart00: according to https://jsoup.org/apidocs/org/jsoup/nodes/Attributes.html , I suspect the `.removeAttr("VALUE")` isn't necessary

bstewart00: Maybe include different casing in the test cases to make sure. I'm surprised this wouldn't also apply to the other branches. Aren't we using the same version on all of them?

zeyuwu: I confirmed for `jsoup` used for Java 7 and above, when setting the attribute the logic is implemented as `attributes().putIgnoreCase(attributeKey, attributeValue);`, while for Java 6 it is `attributes.put(attributeKey, attributeValue);`. So in Java6 it is not case-insensitive.

zeyuwu: jsoup verisons: java6: 1.10.3, java7: 1.11.3, java8: 1.11.3

bstewart00: If 1.11 is also supported by java6, I wouldn't be opposed to simply upgrading the package.

zeyuwu: `Updated language level to Java 7 from Java 5.` Nah we can't ;-;

bstewart00: Ok

bstewart00: although this seems a bit useless when we only check `element.hasAttr("value")`. I think we can get away with just handling `value` lowercase tbh

zeyuwu: The reason hasAttr() is "value" only because hasAttr is always case-insensitive, so by using hasAttr we never know if the actual thing is "value" or "VALUE". The double removal of "value" and "VALUE" is just so that we don't accidentally end up with two values in the resulting html.

bstewart00 (approved):

Soichiro725 (approved):

PR #168, opened 2020-11-09
Revert: Add TLS 1.2 support for java 6

bstewart00:
Revert https://github.com/WOVNio/wovnjava/pull/159

The code is basically correct in how we are installing BouncyCastle, but there are several environment problems we weren't aware of.

```
Caused by: java.security.NoSuchAlgorithmException: Algorithm ECDH not available
        at javax.crypto.KeyAgreement.getInstance(DashoA13*..)
```

```
                                 com.github.wovnio.wovnjava.ApiException: Exception : unable to create JcaTlsCrypto: SecureRandom DEFAULT implementation not found: java.lang.ClassNotFoundException: org.bouncycastle.jcajce.provider.drbg.DRBG$Default
        at com.github.wovnio.wovnjava.Api.translate(Api.java:69)
```

For now I am reverting this change until we understand these issues more. I suspect there is a problem with the fat jar packaging we are doing.

https://stackoverflow.com/questions/46175454/exclude-bouncycastle-jar-but-include-everything-else-in-a-fat-jar
https://github.com/bcgit/bc-java/issues/514

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
-    <version>1.0.3-jdk6</version>
+    <version>1.0.4-jdk6</version>
     <url>https://github.com/WOVNio/wovnjava</url>
 
     <licenses>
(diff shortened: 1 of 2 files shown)


iwai-wovn (approved):
The customer seems to have worked, with this fix.
Thanks you, quick response!!

Soichiro725 (approved):

PR #169, opened 2020-12-22
[Feature] Add insert doc lang feature.

zeyuwu:
https://wovnio.atlassian.net/browse/ST-599

When an HTML is converted, if the `<html>` tag doesn't have a `lang` attribute, the default language will be added.

Before: `<html><head>...`
After: `<html lang="en"><head>...`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
-    <version>1.5.1</version>
+    <version>1.6.0</version>
     <url>https://github.com/WOVNio/wovnjava</url>
 
     <licenses>
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
@@ -32,7 +32,7 @@ class HtmlConverter {
         removeWovnIgnore();
         removeClassIgnore();
         removeForm();
-        insertDocLangIfEmpty();
+        insertHtmlLangAttribute();
         return doc.html();
     }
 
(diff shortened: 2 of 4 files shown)


aw-leigh (approved):
Looks good, that was simpler than I expected.

Soichiro725:
You can copy test cases from other libraries.
https://github.com/WOVNio/Wovn.cs/blob/ccd96baa01aa9639f4cce84f5aa3738aeb33e2e3/wovn.cs.Tests/Html/HtmlConverterTest.cs#L945

If same method names with other libraries are used, it it more easy to understand.
- insertDocLangIfEmpty -> insertHtmlLangAttribute

Soichiro725:
Please update version.

zeyuwu:
> If same method names with other libraries are used, it it more easy to understand.

This is a good point.

> You can copy test cases from other libraries.

The other libraries have tests focused on the correct parsing of the `lang` attribute (e.g. special cases where single quotes are used or cases where the attribute values containing `-`) because they involve string-based HTML manipulations. This is not the case in wovnjava, therefore the test cases are simplified. If we were to test for those cases in wovnjava, we'd effectively be testing for the correctness of the third-party HTML parsing library, which is redundant.

Soichiro725:
> The other libraries have tests focused on the correct parsing of the lang attribute (e.g. special cases where single quotes are used or cases where the attribute values containing -) because they involve string-based HTML manipulations. This is not the case in wovnjava, therefore the test cases are simplified. If we were to test for those cases in wovnjava, we'd effectively be testing for the correctness of the third-party HTML parsing library, which is redundant.

You are right. For Java, test case related to quotes are not necessary.

Soichiro725:
Looks good.

Soichiro725 (approved):

PR #170, opened 2021-04-08
Check for widget versioning snippets

bstewart00:
https://wovnio.atlassian.net/browse/ST-826

Snippets like `<script src="https://wovn.global.ssl.fastly.net/widget/abcdef">` should also be considered a widget snippet and removed.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
@@ -18,6 +18,12 @@ class HtmlConverter {
     private final HashMap<String, String> hreflangMap;
     private final HtmlReplaceMarker htmlReplaceMarker;
 
+    private static final String[] WOVN_WIDGET_URLS = new String[] {
+        "j.wovn.io",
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -18,6 +18,12 @@
     private final HashMap<String, String> hreflangMap;
     private final HtmlReplaceMarker htmlReplaceMarker;
 
+    private static final String[] WOVN_WIDGET_URLS = new String[] {
+        "j.wovn.io",
+        "j.dev-wovn.io:3000",
+        Settings.DefaultApiUrlBase
```
Soichiro725, line 24: I think this condition is enough.
But, WOVN.cs and Proxy have `/widget` to `https://wovn.global.ssl.fastly.net`.

Soichiro725 (approved):
Looks good.

Soichiro725 (approved):
Looks good

zeyuwu (approved):
looks good.

PR #171, opened 2021-04-28
feature: Add outboundProxyHost/Port settings to enable communication with html swapper via a proxy.

bstewart00:
https://wovnio.atlassian.net/browse/ST-881

version 1.2.0-java7

Add outboundProxyHost/Port settings to enable communication with html swapper via a proxy. 

Unfortunately we already have a `useProxy` setting which is used for a different purpose (when the wovnjava request is coming from a reverse proxy and doesn't reflect the true browser URL). I've tried to be explicit that this setting is for proxying outbound requests.

```
<init-param>
  <param-name>outboundProxyHost</param-name>
  <param-value>proxy.company.com</param-value>
</init-param>
<init-param>
  <param-name>outboundProxyPort</param-name>
  <param-value>8080</param-value>
</init-param>
```


zeyuwu (approved):
Looks good. But have we confirmed that the proxy client is using is HTTP and not SOCKS or some other type?

bstewart00:
> Looks good. But have we confirmed that the proxy client is using is HTTP and not SOCKS or some other type?

No, i'm just making an assumption. Who uses SOCKS? 😛 I guess we should check though.

iwai-wovn (approved):

bstewart00:
We can make another release later if we need SOCKS support.

PR #174, opened 2021-05-24
Fix missing hreflang

zeyuwu:
https://wovnio.atlassian.net/browse/ST-898

Make wovnjava add hreflangs before sending non-source language HTML to html-swapper.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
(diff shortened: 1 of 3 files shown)


Soichiro725 (approved):

Soichiro725:
Looks good. But it needs to fix tests.

bstewart00 (approved):
This brings the behavior in line with the other backends, but I think potentially we should re-evaluate whether we should be sending hreflangs to html swapper at all, given that it will generate them itself.

bstewart00:
change patch number in pom.xml

PR #177, opened 2021-06-18
[ST-983] Allow explicit encoding setting

zeyuwu:
https://wovnio.atlassian.net/browse/ST-983

Added a new setting item called `encoding`, when set, wovnjava will use that encoding to handle all strings. If not set, wovnjava will attempt to determine the encoding from a predetermined list of encodings.

### DIFF
--- a/README.en.md
+++ b/README.en.md
@@ -116,6 +116,7 @@ sitePrefixPath            |          |
 langCodeAliases           |          |
 customDomainLangs         |          |
 debugMode                 |          | false
+encoding | | 
 
 ### 2.1. projectToken (required)
 
@@ -422,6 +423,10 @@ Using `wovnDebugMode` as a query parameter will activate embedded debug informat
 This is intended to better understand what the problem is if something is not working correctly with wovnjava on your server.
 
 _Note that `wovnCacheDisable` and `wovnDebugMode` is only available when debugMode is turned on in your wovnjava configuration._
+
+### 2.15. encoding
+
+You can choose to optionally explicitly specify the encoding used by your HTML. If `encoding` is not set, wovnjava will attempt to automatically detect the encoding. 
 
 ## Supported Langauges
 
--- a/README.en.md
+++ b/README.en.md
@@ -427,7 +427,8 @@ _Note that `wovnCacheDisable` and `wovnDebugMode` is only available when debugMo
 ### 2.15. encoding
 
 You can choose to optionally explicitly specify the encoding used by your HTML. If `encoding` is not set, wovnjava will attempt to automatically detect the encoding.
-Commonly used encodings are `utf-8`, `Shift_JIS` and `EUC-JP`.
+
+Commonly used encodings are `utf-8`, `Shift_JIS` and `EUC-JP`. For a complete list of supported encodings, please refer to Java's official documentation [here](https://docs.oracle.com/javase/8/docs/technotes/guides/intl/encoding.doc.html).
 
 ## Supported Langauges
 
--- a/README.en.md
+++ b/README.en.md
@@ -426,7 +426,8 @@ _Note that `wovnCacheDisable` and `wovnDebugMode` is only available when debugMo
 
 ### 2.15. encoding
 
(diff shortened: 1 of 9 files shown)


src/test/java/com/github/wovnio/wovnjava/Utf8Test.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class Utf8Test extends TestCase {
+    public void testDefaultEncodingUTF8() {
+        Utf8 unicodeConverter = new Utf8("");
+        String data = new String("Unicøde Sentence　東京 الخط العربي في يونيكود");
+        byte[] rawData = null;
+        try {
+            rawData = data.getBytes("utf-8");
+        } catch (UnsupportedEncodingException e) {
+            System.exit(1);
```
bstewart00, line 19: do you really need to exit inside of a unit test?

src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java
```diff
@@ -16,11 +16,13 @@
...
     private ServletOutputStream output;
     private Headers headers;
+    private Utf8 unicodeConverter;
 
-    WovnHttpServletResponse(HttpServletResponse response, Headers headers) {
+    WovnHttpServletResponse(HttpServletResponse response, Headers headers, String encoding) {
         super(response);
         this.buff = new ByteArrayOutputStream();
         this.headers = headers;
+        this.unicodeConverter = new Utf8(encoding);
```
bstewart00, line 25: you may as well promote this to a constructor parameter. No point passing in `encoding` just to construct `Utf8` out of it. I wouldn't be against renaming it to UnicodeConverter either.

src/test/java/com/github/wovnio/wovnjava/Utf8Test.java
zeyuwu: I suppose I can rethrow the exception... just thought it might not be that clean...

src/main/java/com/github/wovnio/wovnjava/WovnHttpServletResponse.java
zeyuwu: Good point.

pom.xml
```diff
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
-    <version>1.7.2</version>
+    <version>1.7.3</version>
```
bstewart00, line 10: since this is a new feature, a minor bump to 1.8.0 would be better

README.en.md
```diff
@@ -423,6 +424,10 @@ This is intended to better understand what the problem is if something is not wo
 
 _Note that `wovnCacheDisable` and `wovnDebugMode` is only available when debugMode is turned on in your wovnjava configuration._
 
+### 2.15. encoding
+
+You can choose to optionally explicitly specify the encoding used by your HTML. If `encoding` is not set, wovnjava will attempt to automatically detect the encoding. 
+
```
Soichiro725, line 430: It is easier to understand there are candidates like UTF-8, Shift_JIS, EUC-JP, etc.

zeyuwu: Let me add a few common examples.

bstewart00: Linking to (oracle) java documentation about the standard list of encodings might be helpful. https://docs.oracle.com/javase/8/docs/technotes/guides/intl/encoding.doc.html

Soichiro725:
I tried this PR with EUC-JP page in my local.
But, I saw garbled texts.

zeyuwu:
> I tried this PR with EUC-JP page in my local.
> But, I saw garbled texts.

I tried with Shift_JIS, and it works as intended. I'll try to see if EUC-JP is an exception.

Soichiro725:
> I tried with Shift_JIS, and it works as intended.
For me, `Shift_JIS` works, but `EUC-JP` doesn't work.

Soichiro725 (approved):
Looks good.

bstewart00 (approved):

PR #179, opened 2021-06-28
[ST-939] Improve logging

zeyuwu:
https://wovnio.atlassian.net/browse/ST-939

This PR made various improvements to the logging of wovnjava:
- there's a config item `enableLogging`, which takes boolean values, that will determine if wovnjava should log or not
- now the logging make use of `java.util.logging` for maximum comaptibility
- the logs will have a UUID
- the same UUID is sent to html-swapper so the logs are in sync


src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -50,12 +51,16 @@ String translate(String lang, String html) throws ApiException {
             con.setReadTimeout(settings.readTimeout);
             return translate(lang, html, con);
         } catch (UnsupportedEncodingException e) {
+            WovnLogger.log("API error: unsupported encoding.");
             throw new ApiException("UnsupportedEncodingException", e.getMessage());
         } catch (MalformedURLException e) {
+            WovnLogger.log("API error: malformed URL.");
             throw new ApiException("MalformedURLException", e.getMessage());
         } catch (IOException e) {
+            WovnLogger.log("API error: IO exception.");
```
bstewart00, line 60: Losing information from the exception here. Does passing in `e` to the log message include the original exception message?

src/main/java/com/github/wovnio/wovnjava/Headers.java
```diff
@@ -29,14 +29,20 @@
...
-        this.urlLanguagePatternHandler = urlLanguagePatternHandler;
-
+        this.urlLanguagePatternHandler = urlLanguagePatternHandler;            
         String clientRequestUrl = UrlResolver.computeClientRequestUrl(request, settings);
         this.requestLang = this.urlLanguagePatternHandler.getLang(clientRequestUrl);
         this.clientRequestUrlInDefaultLanguage = this.urlLanguagePatternHandler.convertToDefaultLanguage(clientRequestUrl);
+        if (this.requestLang == null) {
+            WovnLogger.log("Request cannot be determined");
+        } else {
+            WovnLogger.log(String.format("Request language is: %s", this.requestLang.code));
```
bstewart00, line 39: Would maybe try and avoid the if block here to be shorter:
```
WovnLogger.log(String.format("Request language is: %s", this.requestLang != null ? this.requestLang.code : "null"));
```

src/main/java/com/github/wovnio/wovnjava/WovnLogger.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+class WovnLogger {
+    private static Logger logger = Logger.getLogger("wovnLogger");
+    private static boolean enabled = false;
+    private static String uuid = "NO_UUID";
+    private final static String prefix = "WOVN";
+
+    public static void enbale() {
```
bstewart00, line 13: typo

src/main/java/com/github/wovnio/wovnjava/Api.java
zeyuwu: I didn't include the exception because they are thrown right after that and a stack trace would be in the log. 
If e is included, it will added a stack trace right after the error message.
Now I think about it I don't even think these log messages are needed, because the stack trace would be logged anyway.

bstewart00 (approved):
Works much better!

a-bougouin (approved):
> - the logs will have a UUID
> - the same UUID is sent to html-swapper so the logs are in sync

That's nice :+1:

PR #182, opened 2021-07-14
Fixed incorrect string quality check

zeyuwu:
https://wovnio.atlassian.net/browse/ST-940

Turns out it's just a terrible string equality check error. The setHeader method don't see any matching header, and therefore is not translating anything.

Included is also a minor change in the makefile. Added a new constant that represents wovnjava's version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ SHELL        := /bin/bash
 .SHELLFLAGS  := -eu -o pipefail -c
 
 VERSION := 8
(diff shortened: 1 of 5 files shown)


bstewart00:
update patch number. And we definitely need tests. Either testing WovnHttpServletResponse directly or through WovnServletFilterTest maybe.

bstewart00 (approved):

Soichiro725 (approved):
Thank you for changing makefile.
Looks good.

PR #183, opened 2021-07-20
Update ignore file extension

zeyuwu:
Changes:
- added two new file extensions into the list of "should not translate"
- changed how we detect if a file is binary/needs UTF-8 conversion treatment
  - before: we load the entire file and see if we can convert it to a string (this doesn't work, as you always get _some_ string back)
  - after: we look at the content-type, only those that have `text/` in them are not binary + needs UTF-8 conversion.

Other changes:
- updated make file to make stale build artifacts less likely to exist

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
@@ -4,6 +4,10 @@ import org.jsoup.Jsoup;
 import org.jsoup.nodes.Element;
 
 class HtmlChecker {
+    public boolean isTextFileContentType(String contentType) {
+        return contentType == null || contentType.toLowerCase().contains("text/") || contentType.toLowerCase().contains("html");
+    }
+
     public boolean canTranslate(String contentType, String html) {
         return canTranslateContentType(contentType) &&
             canTranslateContent(html);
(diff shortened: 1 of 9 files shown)


src/main/java/com/github/wovnio/wovnjava/HtmlChecker.java
```diff
@@ -4,6 +4,10 @@
 import org.jsoup.nodes.Element;
 
 class HtmlChecker {
+    public boolean isTextFileContentType(String contentType) {
```
bstewart00, line 7: I don't think there is a need for this. Just use `canTranslateContentType`. I think we should remove `canTranslate` and call the parts individually

bstewart00: Or is it not possible for non-html text responses to go through the binary branch?

zeyuwu: There's a subtle difference between "translatable" and "non-binary". For example, CSS files are not translatable, but they might have `::before` stuff in Shift_JIS.

bstewart00 (approved):
looks good. I assume the binary branch actually works if taken though?

iwai-wovn:
@zeyuwu 
Customer is using java7, so please merge into the java7 branch as well.

Soichiro725 (approved):
Looks good

iwai-wovn (approved):

PR #185, opened 2021-07-29
Added Support for Universal Routes in Docker Env

zeyuwu:
This PR is a quality of life PR that improves debugging using the docker environment. A new filter is provided in the java8 docker example website.

Most notably, this PR adds the ability to send a request to any route prefixed by `/custom_response`, and dictate what the response should be.

You can decide:

- Response headers
- Response HTTP status code
- Response body
- You need to include a JSON body with such requests:
```
{
    "response": {
        "body": "foo2",
        "content-type": "text/html",
        "status": 200,
        "headers": { "X-WovnTest": "Foo" }
    }
}
```

Theoretically speaking, content-type is also a header, but it is included in its separate field.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ SHELL        := /bin/bash
 .SHELLFLAGS  := -eu -o pipefail -c
 
 VERSION := 8
-WOVN_VERSION := 1.9.3
+WOVN_VERSION := 1.9.2
 TARGET_DIR = ${PWD}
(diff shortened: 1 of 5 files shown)


bstewart00 (approved):
don't need to increment the version here I think since it is just test/docker changes. looks good though

Soichiro725 (approved):
Looks good.

PR #186, opened 2021-08-13
[SUP-969] Add Handling of 422 API Error

zeyuwu:
https://wovnio.atlassian.net/browse/SUP-969

With this PR, an HTTP 422 status code returned from html-swapper no longer results in an unhandled exception being thrown.

- Added new exception `ApiNoPageDataException`, thrown when the html-swapper response code is 422
- The `ApiNoPageDataException` is caught and the request is processed the same way as other failed swaps
- `X-Wovn-Api-Status: ApiDataNotAvailable` header is added to all responses where `ApiNoPageDataException` is caught
- Fixed misc. code warnings (unused variables etc)

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Interceptor.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Interceptor.java
@@ -1,8 +1,6 @@
 package com.github.wovnio.wovnjava;
 
 class Interceptor {
-    public static final String NO_API_DATA_STATUS = "ApiDataNotAvailable";
-
     private final Settings settings;
     private final Headers headers;
     private final Api api;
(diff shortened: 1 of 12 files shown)


src/main/java/com/github/wovnio/wovnjava/Interceptor.java
```diff
@@ -29,6 +31,9 @@ private String apiTranslate(String lang, String body) {
             String translatedBody = api.translate(lang, convertedBody);
             responseHeaders.setApiStatus("Success");
             return converter.restore(translatedBody);
+        } catch (ApiNoPageDataException e) {
+            responseHeaders.setApiStatus(Interceptor.NO_API_DATA_STATUS);
```
Soichiro725, line 35: How about defining an error name at `ApiNoPageDataException.java` and writing it like `e.getName()` to get it?
I don't think `Interceptor.java` needs to know the name of the error.

zeyuwu: That's also good. Let me implement that.

iwai-wovn (approved):

Soichiro725 (approved):
LGTM

PR #188, opened 2021-09-15
[ST-1104] Added support for widget versioning url.

zeyuwu:
Added support for widget versioning.
- added a required option `snippetUrl` that should be set to the URL of the snippet.
  -  if not set, `https://j.wovn.io/1` will be used

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Settings.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Settings.java
@@ -2,6 +2,7 @@ package com.github.wovnio.wovnjava;
 
 import java.util.ArrayList;
 import java.util.Map;
+import java.net.MalformedURLException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
@@ -20,14 +21,12 @@ class Settings {
     public static final String DefaultSnippetUrlProduction  = "//j.wovn.io/1";
     public static final String DefaultSnippetUrlDevelopment = "//j.dev-wovn.io:3000/1";
 
-    // Derived default configuration values
-    public final String defaultVersionedWidgetUrlProduction;
-
     // Required settings
     public final String projectToken;
     public final String urlPattern;
     public final Lang defaultLang;
     public final ArrayList<Lang> supportedLangs;
+    public final String snippetUrl;
 
     // Optional settings
     public final boolean devMode;
@@ -40,7 +39,6 @@ class Settings {
     public final Map<Lang, String> langCodeAliases;
     public final CustomDomainLanguages customDomainLanguages;
 
-    public final String snippetUrl;
     public final String apiUrl;
     public final String originalUrlHeader;
     public final String originalQueryStringHeader;
@@ -56,8 +54,6 @@ class Settings {
 
     public final String encoding;
 
-    public final boolean enableWidgetVersioning;
-
     Settings(FilterConfig config) throws ConfigurationError {
         FilterConfigReader reader = new FilterConfigReader(config);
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -14,8 +14,8 @@
     // Default configuration values
     public static final int DefaultTimeout = 1000;
     public static final String DefaultApiUrlBase  = "https://wovn.global.ssl.fastly.net";
+    public static final String DefaultWidgetVersionUrlPrefix = "https://cdn.wovn.io/widget/";
```
bstewart00, line 17: Ok so wovnjava already knew about versioned snippet URLs for the purposes of removing snippets from the page. So I think we should keep this behavior, in addition to adding the snippetUrl setting.

The previous variable was more correct though, use that 
`public static final String DefaultVersionedWidgetUrlProduction = DefaultApiUrlBase + "/widget";`

src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -14,8 +14,8 @@
...
     public static final String DefaultApiUrlProduction  = DefaultApiUrlBase + "/v0/";
-    public static final String DefaultVersionedWidgetUrlProduction  = DefaultApiUrlBase + "/widget";
     public static final String DefaultApiUrlDevelopment = "http://localhost:3001/v0/";
     public static final String DefaultSnippetUrlProduction  = "//j.wovn.io/1";
```
bstewart00, line 20: change this to `https://j.wovn.io/1`. I don't want to further encourage usage of `//`

src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -129,6 +129,13 @@ private Lang verifyDefaultLang(String value) throws ConfigurationError {
         return lang;
     }
 
+    private String verifySnippetUrl(String value) throws ConfigurationError {
```
bstewart00, line 132: this is an opt-in setting and we don't want to affect existing users. So default to the production url instead.

src/main/java/com/github/wovnio/wovnjava/Settings.java
zeyuwu: This one is just simply unused. When removing snippet URL the source comes from the `snippetUrl` defined in the setting.

bstewart00: OK I guess that's fine. Basically we're going to trust that if they are using versioning, they are going to use this setting

src/main/java/com/github/wovnio/wovnjava/Settings.java
bstewart00: use the `stringOrDefault` method. `verify` methods should be ones that can throw

src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -79,7 +78,6 @@
 
         String defaultApiUrl = this.devMode ? DefaultApiUrlDevelopment : DefaultApiUrlProduction;
         this.apiUrl = stringOrDefault(reader.getStringParameter("apiUrl"), defaultApiUrl);
-        this.snippetUrl = this.devMode ? DefaultSnippetUrlDevelopment : DefaultSnippetUrlProduction;
```
bstewart00, line 82: I think you want to preserve this behavior

zeyuwu: I'm not sure it's _that_ useful but let's get it back.

bstewart00 (approved):
looks good

Soichiro725 (approved):
Looks good.

PR #189, opened 2021-10-11
feature: [ST-1100] Add compressApiRequests setting, defaulted to true.

bstewart00:
wovnio.atlassian.net/browse/ST-1100

We currently gzip the request to html swapper. There is now a compress_api_request setting that can be set to false to disable this.

This is only to make it easier to use MockServer to assert the incoming request as JSON in integration tests.

Small refactor to stop constructing a url encoded string manually and generate it from a Map instead.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Api.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Api.java
@@ -14,7 +14,6 @@ import java.net.Proxy;
 import java.net.SocketTimeoutException;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.security.MessageDigest;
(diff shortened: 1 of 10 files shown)


src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -14,7 +14,9 @@
 import java.net.SocketTimeoutException;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
+import java.util.HashMap;
```
zeyuwu, line 17: unused import

src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -72,15 +74,27 @@ String translate(String lang, String html) throws ApiException, ApiNoPageDataExc
...
+
+            Map<String, String> apiParams = getApiParameters(lang, html);
+            String urlEncodedBody =  FormUrlEncoding.encode(apiParams);
+            byte[] apiBodyBytes = urlEncodedBody.getBytes();
+
+            if (this.settings.compressApiRequests) {
+                ByteArrayOutputStream compressedBody = gzipStream(apiBodyBytes);
+                compressedBody.writeTo(out);
+                con.setRequestProperty("Content-Type", "application/octet-stream");
+                con.setRequestProperty("Content-Length", String.valueOf(compressedBody.size()));
```
zeyuwu, line 87: Just for general correctness I think we should also include `Content-Encoding: gzip` here.

zeyuwu (approved):
Looks good.

aw-leigh (approved):

PR #194, opened 2021-11-10
fix: rename ref="alternate" -> rel="alternate" for java6

iwai-wovn:
I'll cherry-pick from java7 fix.


zeyuwu (approved):
I guess we can ignore those failed tests for now since this is only java6

Soichiro725 (approved):
Looks good.

PR #195, opened 2021-12-09
Add JSON support

zeyuwu:
https://wovnio.atlassian.net/browse/ST-1216

`simple_json` was used to serialize the request body because it produces a more "sane" serialized string. (Only down side is that it also escapes `/` as `\/`)
When serializing with `jsonic` too many characters get unicode escaped and it's very unreadable.


aw-leigh (approved):
I'd hate to see the less-sane escaped string!

PR #197, opened 2022-01-04
[ST-1251] Provide a port and host override for WOVN.java

zeyuwu:
https://wovnio.atlassian.net/browse/ST-1251

- added support for fixedScheme, fixedPort and fixedHost options.

Hold until https://github.com/WOVNio/html-swapper/pull/697 is deployed.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Settings.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Settings.java
@@ -56,6 +56,7 @@ class Settings {
     public final String fixedHost;
     public final String fixedScheme;
     public final int fixedPort;
+    public final boolean hasUrlOverride;
 
     Settings(FilterConfig config) throws ConfigurationError {
         FilterConfigReader reader = new FilterConfigReader(config);
@@ -105,10 +106,14 @@ class Settings {
         this.fixedScheme = stringOrDefault(reader.getStringParameter("fixedScheme"), "");
         this.fixedPort = positiveIntOrDefault(reader.getIntParameter("fixedPort"), -1);
         this.verifyFixedURLConfigs(this.fixedHost, this.fixedScheme, this.fixedPort);
+        this.hasUrlOverride = this.fixedPort != -1;
     }
 
     private void verifyFixedURLConfigs(String fixedHost, String fixedScheme, int fixedPort) throws ConfigurationError {
-        if ( (fixedHost.isEmpty() != fixedScheme.isEmpty()) || (fixedScheme.isEmpty() != (fixedPort == -1))) {
+        // all three settings are either all set, or none is set
+        Boolean allTheSame = fixedHost.isEmpty() == fixedScheme.isEmpty() && fixedScheme.isEmpty() == (fixedPort == -1);
+
+        if (!allTheSame) {
             throw new ConfigurationError("Missing configuration: \"fixedHost\", \"fixedScheme\" and \"fixedPort\" must all be defined");
         }
     }
--- a/src/main/java/com/github/wovnio/wovnjava/Settings.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Settings.java
@@ -53,6 +53,10 @@ class Settings {
 
     public final String encoding;
 
+    public final String fixedHost;
+    public final String fixedScheme;
+    public final String fixedPort;
+
     Settings(FilterConfig config) throws ConfigurationError {
         FilterConfigReader reader = new FilterConfigReader(config);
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/github/wovnio/wovnjava/UrlResolver.java
```diff
@@ -7,16 +7,21 @@ private UrlResolver() {}
 
     static String computeClientRequestUrl(HttpServletRequest request, Settings settings) {
         String scheme = request.getScheme();
-        String host = clientRequestHostAndPort(request, settings.useProxy);
+        
+        if (!settings.fixedScheme.isEmpty()) {
+            scheme = settings.fixedScheme;
+        }
+
+        String host = clientRequestHostAndPort(request, settings);
```
bstewart00, line 15: I think this reads better if you structure it like (loosely, this isn't valid java)
```
String host = overrideHost() || clientRequestHostAndPort(request, settings)
String scheme = settings.fixedScheme || request.getScheme()
```

That way `clientRequestHostAndPort` is really about reading the data from the actual request, and is separated from the overrides

src/main/java/com/github/wovnio/wovnjava/UrlResolver.java
```diff
@@ -25,6 +30,15 @@ private static String clientRequestHostAndPort(HttpServletRequest request, boole
                 port = Integer.parseInt(forwardedPort);
             }
         }
+
+        if (!settings.fixedPort.isEmpty()) {
+            port = Integer.parseInt(settings.fixedPort);
```
bstewart00, line 35: handle the parsing in Settings, don't expose it to the rest of wovnjava

src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -96,6 +100,9 @@
         this.outboundProxyPort = reader.getIntParameter("outboundProxyPort");
 
         this.encoding = stringOrDefault(reader.getStringParameter("encoding"), "");
+        this.fixedHost = stringOrDefault(reader.getStringParameter("fixedHost"), "");
+        this.fixedScheme = stringOrDefault(reader.getStringParameter("fixedScheme"), "");
+        this.fixedPort = stringOrDefault(reader.getStringParameter("fixedPort"), "");
```
bstewart00, line 105: use `positiveIntOrDefault`. Also I want to validate that if you specify one of these settings, you must specify all of them, even in the port number.

zeyuwu: Oh. Will add that.

bstewart00:
You can look at https://github.com/WOVNio/Wovn.cs/blob/master/README.md#27-fixed_scheme-fixed_host-fixed_port for inspiration for the readme

src/main/java/com/github/wovnio/wovnjava/Settings.java
```diff
@@ -96,6 +100,17 @@
...
         this.encoding = stringOrDefault(reader.getStringParameter("encoding"), "");
+
+        this.fixedHost = stringOrDefault(reader.getStringParameter("fixedHost"), "");
+        this.fixedScheme = stringOrDefault(reader.getStringParameter("fixedScheme"), "");
+        this.fixedPort = positiveIntOrDefault(reader.getIntParameter("fixedPort"), -1);
+        this.verifyFixedURLConfigs(this.fixedHost, this.fixedScheme, this.fixedPort);
+    }
+
+    private void verifyFixedURLConfigs(String fixedHost, String fixedScheme, int fixedPort) throws ConfigurationError {
+        if ( (fixedHost.isEmpty() != fixedScheme.isEmpty()) || (fixedScheme.isEmpty() != (fixedPort == -1))) {
```
bstewart00, line 111: This looks very confusing and I have to think about it a lot. Why are we comparing these against each other? I think there should be a better way to write it.

bstewart00: For example
1. Check if all parameters are empty or -1, return true if so
2. Otherwise assert `notEmpty && notEmpt && port != -1`

zeyuwu: We want to test if three things a, b, c are the same. The most efficient form would be `a == b && b == c`. And to negate it it becomes `a != b || b != c`. 

I think I can use the `a == b && b == c` form and do an early return if it's the case. That should be simpler. What do you think?

bstewart00: we should optimize for readability here, we don't need to worry about a few extra boolean comparisons. Extract variables if you want, the compiler doesn't care if we split it up over multiple lines.

bstewart00: the way you've written it is more verbose than it could be, which isn't helping. If you want to assert three variables are equal you can do `a == b == c` and not need conditionals

zeyuwu: `a == b == c` wouldn't work would it, because it evaluates from left to right. `false == false == true` would be `true`.

bstewart00: ok yeah, but this is exactly why we shouldn't make code too clever for the sake of efficiency and have to think about what it means 😛

bstewart00: OK I like it now. Just having an explicit variable name is helpful.

bstewart00 (approved):
minor version bump and it is good to go. Good tests.

aw-leigh (approved):
Nice work

PR #200, opened 2022-01-25
Fix global proxy support java8

zeyuwu:
https://wovnio.atlassian.net/browse/SUP-1257

If Proxy.NO_PROXY is set it will override any existing global proxy settings.


bstewart00 (approved):

Soichiro725:
Related slack thread: https://wovnio.slack.com/archives/CGM9EKNLX/p1643091221071500

Soichiro725 (approved):
Looks good.

PR #205, opened 2022-02-08
feature: [ST-1269] translate canonical url + search engine api timeout settings

bstewart00:
https://wovnio.atlassian.net/browse/ST-1269
1. Add `api_timeout_search_engine_bots` setting. Having a longer timeout for search engines makes it more likely they will receive a translated version of the page.
2. Add `translate_canonical_tag` setting. This setting adds lang codes to `<link rel="canonical" href="">` tags in the fallback snippet HTML. In the event our API request does time out,  it should not negatively impact SEO. This can be set to `false` to disable translating this tag completely (including in html swapper)

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/Api.java
+++ b/src/main/java/com/github/wovnio/wovnjava/Api.java
@@ -51,8 +51,15 @@ class Api {
               Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(this.settings.outboundProxyHost, this.settings.outboundProxyPort));
               con = (HttpURLConnection) url.openConnection(proxy);
             }
-            con.setConnectTimeout(settings.connectTimeout);
-            con.setReadTimeout(settings.readTimeout);
+            int connectTimeout = settings.connectTimeout;
+            int readTimeout = settings.readTimeout;
+            if (this.headers.getIsValidRequest()) {
+                connectTimeout = this.settings.apiTimeoutSearchEngineBots;
+                readTimeout = this.settings.apiTimeoutSearchEngineBots;
+            }
+
+            con.setConnectTimeout(connectTimeout);
+            con.setReadTimeout(readTimeout);
             return translate(lang, html, con);
         } catch (UnsupportedEncodingException e) {
             WovnLogger.log("API error: unsupported encoding.");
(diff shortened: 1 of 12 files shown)


src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -179,6 +184,20 @@ private void appendHrefLang() {
...
 
+    private void translateCanonicalTag(String lang) {
+        Elements elements = doc.head().getElementsByTag("link");
+        for (Element element : elements) {
+            String rel = element.attr("rel");
+            if (rel != null && rel.equalsIgnoreCase("canonical")) {
+                String href = element.attr("href");
+                if (href != null) {
+                    String translatedCanonicalUrl = this.headers.locationWithLangCode(href);
+                    element.attr("href", translatedCanonicalUrl);
```
bstewart00, line 195: a welcome change from the other libraries where we need to regex the html 😌

src/main/java/com/github/wovnio/wovnjava/Api.java
```diff
@@ -51,8 +51,15 @@ String translate(String lang, String html) throws ApiException, ApiNoPageDataExc
...
-            con.setReadTimeout(settings.readTimeout);
+            int connectTimeout = settings.connectTimeout;
+            int readTimeout = settings.readTimeout;
+            if (this.headers.isSearchEngineBot()) {
+                connectTimeout = this.settings.apiTimeoutSearchEngineBots;
+                readTimeout = this.settings.apiTimeoutSearchEngineBots;
+            }
+
+            con.setConnectTimeout(connectTimeout);
+            con.setReadTimeout(readTimeout);
```
bstewart00, line 62: I noticed that wovnjava has separate settings for these two. I'm tempted to deprecate them in favour of a simpler `apiTimeout` setting. I don't think users particularly need to know about the difference between the connect/read timeout.

src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -179,6 +184,20 @@ private void appendHrefLang() {
         }
     }
 
+    private void translateCanonicalTag(String lang) {
+        Elements elements = doc.head().getElementsByTag("link");
```
zeyuwu, line 188: jsoup allows selection by CSS tag like so `Elements links = doc.select("a[rel=\"canonical\"]")` could be simpler.

https://jsoup.org/cookbook/extracting-data/selector-syntax

zeyuwu (approved):
Looks good. And it passes backend integration test.

Soichiro725 (approved):
Looks good.
I confirmed that canonical tag change works with my local environment.

PR #208, opened 2022-06-27
TLS 1.2 implementation

bstewart00:
Taken from https://github.com/a--i--r/TLSSocketFactory

I confirmed that when testing against https://tls-v1-2.badssl.com:1012/ , without these changes you would get a handshake error. Afterwards, it can send a request succcessfully.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,8 +26,8 @@ jobs:
            sudo apt-get update
            sudo apt install -y maven
            cd ~/
-           curl -O https://bootstrap.pypa.io/pip/2.7/get-pip.py
-           python get-pip.py --user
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -4,10 +4,6 @@ workflows:
   version: 2
   test:
     jobs:
-      - oraclejdk8
-      - oraclejdk7
-      - openjdk8
-      - openjdk7
```
bstewart00, line 10: I gave up trying to fix the python 3.7 error. We only care about java6 on this branch anyway.

iwai-wovn (approved):

aw-leigh (approved):

PR #224, opened 2022-10-10
migrate to java 11

alexisgeoffrey:
Fixes https://wovnio.atlassian.net/browse/ST-1224

Changed from javax to jakarta

updated relevant dependencies
added some abstract method implementations in `FilterServletOutputStream.java`


src/main/java/com/github/wovnio/wovnjava/FilterServletOutputStream.java
```diff
@@ -36,4 +37,14 @@ public void close() throws java.io.IOException {
...
+
+    @Override
+    public boolean isReady() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void setWriteListener(WriteListener writeListener) {
+        throw new UnsupportedOperationException();
+    }
```
bstewart00, line 49: i think you would just return true for isReady, and do nothing in `setWriteListener`

aw-leigh (approved):

PR #227, opened 2023-04-10
feature: [ST-2002] Add overrideContentLength setting to enable setting "Content-Length" header.

bstewart00:
https://wovnio.atlassian.net/browse/ST-2002
https://wovnio.atlassian.net/browse/SUP-2099

New setting: `overrideContentLength`. Default false. When true, we will set/replace Content-Length header based on the size of the translated response.

```
<init-param>
    <param-name>overrideContentLength</param-name>
    <param-value>true/false</param-value>
  </init-param>
```

This code was previously removed in https://github.com/WOVNio/wovnjava/pull/209. In some environments, setting Content-Length causes issues so it is now an optional setting.

Improved `WovnServletFilterTest`. For the first time, we can now assert on the response body.

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
@@ -93,6 +93,7 @@ public class WovnServletFilter implements Filter {
         responseHeaders.setApiStatus("Unused");
 
         if (headers.getIsPathInDefaultLanguage()) {
+            WovnLogger.log("Passing to next filter");
             chain.doFilter(wovnRequest, wovnResponse);
         } else {
             String newPath = headers.getCurrentContextUrlInDefaultLanguage().getPath();
@@ -119,6 +120,9 @@ public class WovnServletFilter implements Filter {
                 body = originalBody;
             }
 
+            if (this.settings.overrideContentLength) {
+                wovnResponse.setContentLength(body.getBytes().length);
+            }
             wovnResponse.setCharacterEncoding("UTF-8");
             PrintWriter out = response.getWriter();
             out.write(body);
--- a/src/test/java/com/github/wovnio/wovnjava/FilterChainMock.java
+++ b/src/test/java/com/github/wovnio/wovnjava/FilterChainMock.java
@@ -1,5 +1,7 @@
 package com.github.wovnio.wovnjava;
 
+import java.io.IOException;
+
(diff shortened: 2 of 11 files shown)


iwai-wovn (approved):

src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
```diff
@@ -93,6 +93,7 @@ private void tryTranslate(Headers headers, RequestOptions requestOptions, HttpSe
         responseHeaders.setApiStatus("Unused");
 
         if (headers.getIsPathInDefaultLanguage()) {
+            WovnLogger.log("Passing to next filter");
```
suplo, line 96: we should keep this log?

src/test/java/com/github/wovnio/wovnjava/FilterChainMock.java
```diff
@@ -8,9 +10,18 @@
 public class FilterChainMock implements FilterChain {
     public HttpServletRequest req;
     public ServletResponse res;
+    public String originalResponseBody;
+
+    public FilterChainMock(String originalResponseBody) {
+        this.originalResponseBody = originalResponseBody;
+
```
suplo, line 17: please remove blank line

tim-wovn (approved):
I don't know Java but the changes look simple enough. 🙈 🙇  LGTM!

suplo:
there is this change https://github.com/WOVNio/wovnjava/pull/209/files#diff-cf08d73239cf7a76e812724c982b272df4368e3f1448e5e7ade280dacb3a6992L99, should we revert it too when the flag is true?

src/main/java/com/github/wovnio/wovnjava/WovnServletFilter.java
bstewart00: yes, we otherwise don't have any logging for this branch

bstewart00:
> there is this change [#209 (files)](https://github.com/WOVNio/wovnjava/pull/209/files#diff-cf08d73239cf7a76e812724c982b272df4368e3f1448e5e7ade280dacb3a6992L99), should we revert it too when the flag is true?

this setting only influences the response content-length. html swapper request content-length isn't an issue and maybe I should put that code back 🤷

suplo (approved):
Thanks @bstewart00!

PR #229, opened 2023-06-29
fix: [SUP-2251] `wovn-marker` doesn't restore text properly

bryanzerrudo:
https://wovnio.atlassian.net/browse/SUP-2251

Fixes issue where original texts aren't restored properly due to `wovn-marker` being replaced incorrectly.

`wovn-marker` uses incremental counter to track texts to be restored, however, it's stored to a `HashMap` that doesn't store sorted data (texts should be restored LIFO). As such, it's possible for the following to happen:
```
// Original text
<input type="hidden" name="field_1" value="100">
// ...
<input type="hidden" name="field_19" value="text">

// Value removed text
<input type="hidden" name="field_1" value="wovn-marker-1">
// ...
<input type="hidden" name="field_19" value="wovn-marker-19">

// Restored text
<input type="hidden" name="field_1" value="100">
// ...
<input type="hidden" name="field_19" value="1009">
```
If `wovn-marker-1` is called before `wovn-marker-19`, `wovn-marker-19` will be replaced by the original text from `wovn-marker-1` + `9`.


Soichiro725:
Version up needed for 3 files like other PRs.
- Makefile
- docker/java8/hello/pom.xml
- pom.xml

src/test/java/com/github/wovnio/wovnjava/HtmlConverterTest.java
```diff
@@ -117,6 +117,16 @@ public void testRemoveForm__Sanitize__SingleQuotes() throws ConfigurationError {
...
...type=\"hidden\" name=\"field_11\" value=\"\"><INPUT type=\"hidden\" name=\"field_12\" value=\"0\"><INPUT type=\"hidden\" name=\"field_13\" value=\"\"><INPUT type=\"hidden\" name=\"field_14\" value=\"\"><INPUT type=\"hidden\" name=\"field_15\" value=\"\"><INPUT type=\"hidden\" name=\"field_16\" value=\"\"><INPUT type=\"hidden\" name=\"field_17\" value=\"0\"><INPUT type=\"hidden\" name=\"field_18\" value=\"\"><INPUT type=\"hidden\" name=\"field_19\" value=\"\"><INPUT type=\"hidden\" name=\"field_20\" value=\"\"><INPUT type=\"hidden\" name=\"field_21\" value=\"https://example.com/a\"><INPUT type=\"hidden\" name=\"field_22\" value=\"https://example.com/b\"><INPUT type=\"hidden\" name=\"field_23\" value=\"https://example.com/c\"><INPUT type=\"hidden\" name=\"field_24\" value=\"https://example.com/d\"><INPUT type=\"hidden\" name=\"field_25\" value=\"\"><INPUT type=\"hidden\" name=\"field_26\" value=\"\"><INPUT type=\"hidden\" name=\"field_27\" value=\"\"><INPUT type=\"hidden\" name=\"field_28\" value=\"foo\"><INPUT type=\"hidden\" name=\"field_29\" value=\"\"><INPUT type=\"hidden\" name=\"field_30\" value=\"\"><INPUT type=\"hidden\" name=\"field_31\" value=\"100\"></FORM></BODY></HTML>";
```
bstewart00, line 121: Just do this with a loop to make this more readable
50 input fields, 50 elements with wovn-ignore on them

bryanzerrudo: Fixed in 89689d3daa44b557883bd3b23f4afb742aa2083c.

Adding `<div wovn-ignore>text</div>` makes the test pass for some reason. Though it fails if it's just the inputs.

Soichiro725 (approved):

bstewart00 (approved):

src/test/java/com/github/wovnio/wovnjava/HtmlConverterTest.java
bstewart00: that's weird but OK we can leave just the inputs

PR #235, opened 2024-09-19
feature: [ST-3320] send response status code to html swapper

bstewart00:
https://wovnio.atlassian.net/browse/ST-3320
https://github.com/WOVNio/html-swapper/pull/1044

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
-    <version>2.0.0</version>
+    <version>2.0.1</version>
     <url>https://github.com/WOVNio/wovnjava</url>
 
     <licenses>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -7,7 +7,7 @@
     <groupId>com.github.wovnio</groupId>
     <artifactId>wovnjava</artifactId>
     <name>wovnjava</name>
-    <version>2.0.0</version>
+    <version>2.0.1</version>
```
bryanzerrudo, line 10: Perhaps could use minor version?
```suggestion
    <version>2.1.0</version>
```

I think "sending a new parameter to an API" can be considered more as a feature than a bugfix.

bstewart00: it's not really a public API from wovnjava's perspective

bryanzerrudo: Wordpress ([PR](https://github.com/WOVNio/wordpress_plugin/pull/124)) uses minor version though. I think we should change one or the other.

bstewart00: I don't think that was strictly correct, but for consistency i'll just use minor then

bryanzerrudo (approved):

Soichiro725 (approved):

PR #238, opened 2026-06-30
fix(SUP-4254): use plain-key markers for title and textarea

KendrickVP:
## Summary
- Use plain-key `addValue` markers for `<title>` and `<textarea>` instead of HTML comment markers when stripping `wovn-ignore` content
- RCDATA elements treat inner content as plain text, so `<!--wovn-marker-N-->` becomes a literal value (e.g. leaked into form submissions) instead of a hidden comment
- Matches the WOVN.php pattern for `<title>` and extends the same approach to `<textarea>` (same RCDATA behavior)
- Fixes inquiry form leaking marker strings into auto-reply emails for users

### DIFF
--- a/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
+++ b/src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
@@ -226,9 +226,20 @@ class HtmlConverter {
     }
 
     private void replaceNodeToMarkerComment(Element element) {
+        // RCDATA elements (textarea, title) treat inner content as plain text, not HTML.
+        // HTML comment markers would become literal text and leak into form submissions.
+        if (isRcdataElement(element)) {
+            return;
+        }
+
         String commentKey = htmlReplaceMarker.addCommentValue(htmlReplaceMarker.revert(element.html()));
         element.html("");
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/wovnio/wovnjava/HtmlConverter.java
```diff
@@ -226,9 +226,18 @@ private void insertHtmlLangAttribute() {
     }
 
     private void replaceNodeToMarkerComment(Element element) {
-        String commentKey = htmlReplaceMarker.addCommentValue(htmlReplaceMarker.revert(element.html()));
-        element.html("");
-        element.appendChild(new Comment(commentKey));
+        String tagName = element.tagName();
+        String elementHtml = htmlReplaceMarker.revert(element.html());
```
bstewart00, line 230: This part is weird and doesn't match wovn.php, but it was already doing it so 🤷 (PHP aborts and does nothing if the element already contained the marker. Here we are reverting and re-applying it)

bstewart00 (approved):

tim-wovn (approved):
Thank you!

bryanzerrudo (approved):

